PR #170, opened 2026-04-24
v2: JiraClient enhancements (PR #4)

scottlepp:
## Summary

Layer 1 plumbing before Layer 2 (consolidated tools) and Layer 3 (code-api) go live. Nothing user-visible changes — \`JiraClient\`'s public surface is unchanged.

### Four enhancements

| What | Where | Why |
|---|---|---|
| Undici connection pool | \`src/core/http.ts\` | Keep-alive + shared pool (8 connections) replaces native fetch. Latency drop on back-to-back calls. |
| cloudId disk cache | \`src/core/tenant-cache.ts\` | 24h TTL, host-keyed. v1 hit \`/_edge/tenant_info\` every cold start; now warm starts skip the network. |
| Metadata LRU | \`src/core/lru.ts\` | 5-min TTL, 64-entry cap. Cuts chatter for \`/field\`, \`/issuetype\`, \`/priority\`, \`/status\`, \`/resolution\`. ~60-line, no dep. |
| 429 backoff | \`httpRequest\` | 3 retries, exponential + jitter, honors \`Retry-After\` (seconds or HTTP date). Previously the client just threw on first 429. |

## Design notes

- **Undici Agent is a lazy module-level singleton** — one pool per process, not per client. A \`closeHttpPool()\` export is included for graceful shutdown.
- **Transport is injectable for tests** via \`__setTransportForTests\`. All retry logic is tested against a stub — no real network in unit tests.
- **Metadata LRU is keyed by path + sorted query params**, so paginated or filtered requests don't collide with each other.
- **\`Ref<T>\` / sandbox / trim from PRs #2/#3 are still ground-truth for response shapes.** This PR only touches *transport*; response-shaping (summary + ref) is still PR #7's job.

## Test changes

Added 30 new unit tests across three files. Also refactored \`sandbox.test.ts\` cleanup to remove only its own session dir plus known fixture names, not the whole root — this fixes a cross-file flake I introduced when two test files shared \`rootCacheDir()\`. The previously-flaky \"does not rewrite\" mtime check is replaced with a deterministic content-sentinel assertion.

## Test plan

- [x] 101 / 101 unit tests pass across 5 consecutive runs
- [x] \`npm run build\` clean
- [ ] Reviewer: confirm the metadata cache exclusion list (currently 5 paths) matches your mental model; add/remove in a follow-up if needed.
- [ ] Reviewer: OK to not expire the disk cache on config change? If a user changes \`JIRA_HOST\` the cache key changes (it's keyed by host), so the stale entry just becomes orphaned rather than incorrectly served.

## Next up

- **PR #5**: Streaming attachment downloader (\`src/core/attachments.ts\`)
- **PR #6**: Central operation manifest (\`src/core/manifest.ts\`)

### DIFF
--- a/src/auth/jira-client.ts
+++ b/src/auth/jira-client.ts
@@ -71,12 +71,18 @@ export class JiraClient {
       return;
     }
 
-    // Prevent multiple parallel initializations
+    // Prevent multiple parallel initializations. On rejection, clear
+    // the promise so the next call can retry — otherwise a transient
+    // init failure (network blip, 5xx on tenant_info) would leave the
+    // client permanently broken.
     if (this.initPromise) {
       return this.initPromise;
     }
 
-    this.initPromise = this.fetchCloudId();
+    this.initPromise = this.fetchCloudId().catch((err) => {
+      this.initPromise = null;
+      throw err;
+    });
     return this.initPromise;
   }
 
--- a/src/auth/jira-client.ts
+++ b/src/auth/jira-client.ts
@@ -1,3 +1,6 @@
+import { httpRequest } from "../core/http.js";
+import { TtlLruCache } from "../core/lru.js";
+import { readTenantCache, writeTenantCache } from "../core/tenant-cache.js";
 import { JiraConfig } from "../config.js";
 
 export interface JiraRequestOptions {
@@ -22,10 +25,28 @@ export class JiraApiError extends Error {
   }
 }
 
+// Metadata endpoints whose responses rarely change within a session.
+// GET requests to these paths are served from an in-memory LRU for the
+// TTL. Everything else bypasses the cache.
+const METADATA_PATHS = new Set([
+  "/field",
+  "/issuetype",
+  "/priority",
+  "/status",
+  "/resolution",
+]);
+
+const METADATA_TTL_MS = 5 * 60 * 1000;
+const METADATA_MAX_SIZE = 64;
+
 export class JiraClient {
   private config: JiraConfig;
   private cloudId: string | null = null;
   private initPromise: Promise<void> | null = null;
+  private readonly metadataCache = new TtlLruCache<string, unknown>({
+    maxSize: METADATA_MAX_SIZE,
+    ttlMs: METADATA_TTL_MS,
+  });
 
   constructor(config: JiraConfig) {
     this.config = config;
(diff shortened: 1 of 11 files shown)


src/auth/jira-client.ts
```diff
@@ -60,39 +81,56 @@ export class JiraClient {
   }
 
   /**
-   * Fetch the cloudId from the tenant_info endpoint
-   * This works for both scoped tokens and classic tokens
+   * Fetch the cloudId from the tenant_info endpoint, preferring a
+   * cached value on disk (24h TTL) to avoid the network hop on warm
+   * starts. Works for both scoped tokens and classic tokens.
    */
   private async fetchCloudId(): Promise<void> {
```
scottlepp, line 94: **Blocker: `initPromise` is never cleared after a rejection — the JiraClient becomes permanently broken after a transient init failure.**

After `fetchCloudId()` throws (network error, non-2xx status, invalid JSON), `this.initPromise` is left holding the rejected `Promise<void>`. The next call to `ensureInitialized()` hits the `if (this.initPromise)` guard — truthy because a rejected Promise is still a Promise object — and immediately awaits the same already-rejected promise. The instance is permanently broken with no recovery path for the caller short of constructing a whole new `JiraClient`.

This PR didn't introduce the guard but substantially rewrote the `fetchCloudId` body, so the failure modes are now richer (disk-cache throw → network throw → JSON parse throw). Each new throw path solidifies the permanent-failure risk.

**Fix:** clear the promise on rejection so the next call can retry:

```ts
// in ensureInitialized():
this.initPromise = this.fetchCloudId().catch((err) => {
  this.initPromise = null; // allow retry on next call
  throw err;
});
return this.initPromise;
```

src/core/http.ts
```diff
@@ -0,0 +1,161 @@
...
+): number {
+  if (retryAfterHeader) {
+    const asNumber = Number(retryAfterHeader);
+    if (Number.isFinite(asNumber) && asNumber >= 0) {
+      return Math.min(asNumber * 1000, opts.maxDelayMs);
+    }
+    const asDate = Date.parse(retryAfterHeader);
+    if (!Number.isNaN(asDate)) {
+      const diff = asDate - now;
+      if (diff > 0) return Math.min(diff, opts.maxDelayMs);
```
scottlepp: **Suggestion: a past `Retry-After` HTTP-date silently falls through to exponential backoff instead of retrying immediately.**

When the server sends a `Retry-After` date that is already in the past (clock skew, queued responses, etc.), `diff <= 0` so the `if (diff > 0) return` guard is not taken. Execution falls through to the exponential backoff formula and introduces a spurious delay of up to several seconds, even though the server's intent was "retry now."

A past date should be treated as "retry immediately" (delay = 0):

```ts
const asDate = Date.parse(retryAfterHeader);
if (!Number.isNaN(asDate)) {
  const diff = asDate - now;
  return diff > 0 ? Math.min(diff, opts.maxDelayMs) : 0;
}
```

The existing tests only cover future dates; a test with `now > asDate` would catch this.

scottlepp:
One blocker, one suggestion — everything else looks solid.

**Blocker** — `src/auth/jira-client.ts:79` (`initPromise` never cleared on rejection)
After any failure in `fetchCloudId()` (network error, non-2xx, bad JSON), `this.initPromise` retains the rejected Promise. Every subsequent call to `ensureInitialized()` sees it as truthy and re-awaits the same rejection, so the JiraClient instance becomes permanently broken with no recovery path. Fix: clear `initPromise` in the rejection handler so the next call can retry. See inline comment.

**Suggestion** — `src/core/http.ts:110` (`computeBackoffMs` drops through on past `Retry-After` date)
When `Retry-After` is a valid HTTP-date that is already in the past, `diff <= 0` causes the code to fall through to exponential backoff rather than returning 0 (retry immediately). Clock skew or a queued 429 response could trigger this. One-line fix: `return diff > 0 ? Math.min(diff, opts.maxDelayMs) : 0`. See inline comment.

The rest of the PR is clean: the undici version fix in the second commit resolves the Node 18/20 compatibility issue, the LRU implementation is correct, tenant-cache TTL/validation logic is sound, test coverage is good across all three new modules, and the sandbox.test.ts cleanup refactor properly scopes teardown to avoid cross-file flakes.

src/auth/jira-client.ts
scottlepp: Fixed in bf36102. `ensureInitialized` now wraps `fetchCloudId()` in a `.catch` that clears `this.initPromise = null` before re-throwing. Regression test at `tests/core/jira-client-init.test.ts`: first request gets a 503 on tenant_info, second request succeeds, proves the client recovers.

Thanks for catching this — you're right that the guard existed in v1 but this PR expanded `fetchCloudId` with two new throw paths (cache-read and JSON-parse) that made the trap easier to fall into.

src/core/http.ts
scottlepp: Fixed in bf36102. A past `Retry-After` date now returns 0 (retry immediately) instead of falling through to exponential backoff. New test with `now > asDate` covers the case.

PR #171, opened 2026-04-24
v2: streaming attachment downloader (PR #5)

scottlepp:
## Summary

Lands [src/core/attachments.ts](../blob/pr/v2-attachments/src/core/attachments.ts), the Layer 1 primitive for getting Jira attachment bytes to disk where the agent can actually use them. Not wired into any tool yet — that's PR #7 and PR #8. Everything in this PR is pure plumbing + tests.

## What it does

- **Streams** the response body directly to \`\${sessionCacheDir}/issues/\${key}/attachments/\${sanitized-filename}\` via \`node:stream/promises pipeline\`. No buffering — a 50MB attachment never hits the Node heap.
- **Atomic writes** via a \`.partial\` temp file + rename. A crash mid-stream won't leave a half-file that would pass the idempotency check next time.
- **Idempotency**: if the target file already exists with a size matching what Jira advertised, skip the network entirely.
- **Size verification**: if the downloaded size doesn't match what Jira advertised, throw and remove the file (defends against truncation by a bad proxy).
- **Preview**: for \`text/*\` and \`application/{json,xml,yaml,javascript}\` mimes, returns the first N characters (default 2000). Binary mimes return \`null\` — the agent will use Claude Code's \`Read\` to render images (paying vision tokens only for files it opens).

## Filename sanitization

Exported as [\`sanitizeFilename\`](../blob/pr/v2-attachments/src/core/attachments.ts#L70-L95). Jira allows filenames with characters that would be dangerous as path segments — an attachment named \`../../.ssh/id\` must not escape the session cache dir.

| Input | Output |
|---|---|
| \`../../../etc/passwd\` | \`passwd\` |
| \`foo\\\\bar.txt\` | \`bar.txt\` |
| \`.hidden\` | \`hidden\` |
| \`(empty)\` or \`///\` | \`attachment\` |

Only the basename survives, leading dots are stripped, names >200 chars are truncated while preserving the extension.

## JiraClient surface change

Adds public \`getAuthorizationHeader()\`. The downloader uses this rather than taking a \`JiraClient\` reference — keeps \`attachments.ts\` pure, testable without a full client, and makes it easy for future code-api stubs to call it.

## Test plan

- [x] 21 new unit tests cover sanitization edge cases, streaming happy path, text vs binary preview, idempotent skip, size-mismatch re-download, HTTP errors, mid-stream failures, cleanup of partial files
- [x] 124 / 124 core tests pass across 5 consecutive runs
- [x] \`npm run build\` clean
- [ ] Reviewer: confirm the 200-char filename cap is reasonable for the Jira population you see in practice

## Next up

- **PR #6**: central operation manifest (\`src/core/manifest.ts\`) — the single declaration that Layer 2 and Layer 3 both consume

### DIFF
--- a/src/core/attachments.ts
+++ b/src/core/attachments.ts
@@ -19,6 +19,7 @@
 // possible here because the attachment's stable identifier is its
 // Jira ID, not its bytes.
 
+import { randomBytes } from "node:crypto";
 import { createWriteStream } from "node:fs";
 import * as fs from "node:fs/promises";
 import * as path from "node:path";
@@ -105,7 +106,21 @@ export function sanitizeFilename(name: string): string {
 
 // --- Directory layout --------------------------------------------------
 
+// Jira issue keys are of the form `PROJECT-123` — uppercase project
+// key (letters, digits, underscore; must start with a letter) + hyphen
+// + numeric id. Anything else is rejected rather than being used as a
+// path segment, since path.join() happily resolves `..` and would
+// escape the session cache dir.
+const ISSUE_KEY_PATTERN = /^[A-Z][A-Z0-9_]*-[0-9]+$/;
+
+export function assertValidIssueKey(key: string): void {
+  if (!ISSUE_KEY_PATTERN.test(key)) {
+    throw new Error(`Invalid Jira issue key: ${JSON.stringify(key)}`);
+  }
+}
+
 function attachmentDir(issueKey: string): string {
+  assertValidIssueKey(issueKey);
   return path.join(sessionCacheDir(), "issues", issueKey, "attachments");
 }
 
@@ -116,12 +131,31 @@ const defaultTransport: AttachmentTransport = async (url, init) => {
     method: init.method,
     headers: init.headers,
   });
+  // The response body is a single-consumption stream. Callers that
+  // read it as `body` (for streaming to disk) must not then also call
+  // `bodyText`, and vice versa. Today's callers respect this (success
+  // path only reads `body`, error path only reads `bodyText`), but
+  // this guard makes accidental double-consumption fail loudly rather
+  // than quietly returning garbage if a future caller forgets.
+  let consumed: "none" | "stream" | "text" = "none";
   return {
     statusCode: res.statusCode,
-    // undici's body is already a Node Readable when the response is
-    // consumed as a stream.
-    body: Readable.from(res.body),
-    bodyText: () => res.body.text(),
+    get body(): Readable {
+      if (consumed === "text") {
+        throw new Error("Response body already consumed via bodyText()");
+      }
+      consumed = "stream";
+      return Readable.from(res.body);
+    },
+    bodyText: () => {
+      if (consumed === "stream") {
+        return Promise.reject(
+          new Error("Response body already consumed via stream"),
+        );
+      }
+      consumed = "text";
+      return res.body.text();
+    },
   };
 };
 
@@ -135,13 +169,32 @@ async function extractPreview(
   maxChars: number,
 ): Promise<string | null> {
   if (!TEXT_MIME_PATTERN.test(mimeType)) return null;
+  // Only read the prefix of the file rather than buffering the whole
+  // thing — a multi-MB log/JSON attachment would otherwise allocate
+  // (buffer + UTF-8 string) proportional to its size even though we
+  // only return maxChars of it.
+  //
+  // Worst-case UTF-8 is 4 bytes per char, plus one extra codepoint of
+  // slack so we can detect whether the file extends beyond the limit.
+  const readBytes = maxChars * 4 + 4;
+  let fh;
   try {
-    const buf = await fs.readFile(filePath);
-    const text = buf.toString("utf8");
-    if (text.length <= maxChars) return text;
+    fh = await fs.open(filePath, "r");
+  } catch {
+    return null;
+  }
+  try {
+    const buf = Buffer.allocUnsafe(readBytes);
+    const { bytesRead } = await fh.read(buf, 0, readBytes, 0);
+    const slice = buf.subarray(0, bytesRead);
+    const truncatedRead = bytesRead === readBytes;
+    const text = slice.toString("utf8");
+    if (!truncatedRead && text.length <= maxChars) return text;
     return `${text.slice(0, maxChars)}…`;
   } catch {
     return null;
+  } finally {
+    await fh.close().catch(() => undefined);
   }
 }
 
@@ -191,7 +244,12 @@ export async function downloadAttachment(
   // Stream to disk. Writing to a temp file and renaming keeps the
   // idempotency check honest: a partial file from an interrupted
   // download won't pass the size check on the next call.
-  const tempPath = `${targetPath}.${process.pid}.partial`;
+  //
+  // The suffix includes random bytes so that two concurrent downloads
+  // of the same attachment within a single process (MCP servers
+  // receive concurrent tool calls) don't both pipe into the same temp
+  // file and corrupt each other.
+  const tempPath = `${targetPath}.${process.pid}.${randomBytes(6).toString("hex")}.partial`;
   try {
     await pipeline(res.body, createWriteStream(tempPath));
     await fs.rename(tempPath, targetPath);
--- a/src/core/attachments.ts
+++ b/src/core/attachments.ts
@@ -1,0 +1,231 @@
+// Streaming attachment downloader.
+//
+// Jira returns attachment metadata with a `content` URL the agent can
+// neither download (needs auth) nor render (binary / possibly large).
+// v1 handed that URL back verbatim, which meant agents couldn't
+// actually use attachments.
+//
+// v2 downloads attachments to the session cache dir and returns a
+// local path, which the agent can feed into Claude Code's `Read` tool
+// (for images, the LLM only pays vision tokens for files it actually
+// opens; text files are cheap and useful as quoting material).
+//
+// Streaming: we use undici's `body` as a Node readable stream and
+// pipe it directly to disk, so a 50 MB attachment never hits Node's
+// heap.
+//
+// Idempotency: if the target path already exists and has the right
+// size, we skip the network entirely. Content-addressing isn't
+// possible here because the attachment's stable identifier is its
+// Jira ID, not its bytes.
+
+import { createWriteStream } from "node:fs";
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import { Readable } from "node:stream";
+import { pipeline } from "node:stream/promises";
+
+import { request as undiciRequest } from "undici";
+
+import { sessionCacheDir } from "./sandbox.js";
+
+export interface AttachmentInput {
+  id: string;
+  filename: string;
+  mimeType: string;
+  size: number;
+  contentUrl: string;
+}
+
+export interface AttachmentDownloadOptions {
+  issueKey: string;
+  authorizationHeader: string;
+  // Overridable for tests. The default transport uses undici.
+  transport?: AttachmentTransport;
+  // Cap preview length (characters) for text/json mimes.
+  previewChars?: number;
+}
+
+export interface AttachmentRef {
+  id: string;
+  filename: string;
+  mimeType: string;
+  size: number;
+  path: string;
+  // Populated only for text-like mimes; null for binaries.
+  preview: string | null;
+}
+
+// The subset of the HTTP transport we need: a streamable response.
+export interface AttachmentHttpResponse {
+  statusCode: number;
+  // A Node Readable stream of the body.
+  body: Readable;
+  // Read the body to text (called on error to surface server message).
+  bodyText: () => Promise<string>;
+}
+
+export type AttachmentTransport = (
+  url: string,
+  init: { method: "GET"; headers: Record<string, string> },
+) => Promise<AttachmentHttpResponse>;
+
+const DEFAULT_PREVIEW_CHARS = 2000;
+
+// --- Filename sanitization --------------------------------------------
+
+// The filename is used as a path segment, so we must strip anything
+// that could break out of the intended directory. Jira itself allows
+// a wide range of characters, but an attachment named `../../.ssh/id`
+// must not escape our session cache.
+const UNSAFE_CHARS = /[/\\\x00]/g;
+const LEADING_DOTS = /^\.+/;
+
+export function sanitizeFilename(name: string): string {
+  // Drop any path components the upload might smuggle in.
+  let base = name.split(/[/\\]/).pop() ?? "";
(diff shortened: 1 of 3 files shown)


src/core/attachments.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+  if (base.length === 0) base = "attachment";
+  return base;
+}
+
+// --- Directory layout --------------------------------------------------
+
+function attachmentDir(issueKey: string): string {
+  return path.join(sessionCacheDir(), "issues", issueKey, "attachments");
```
scottlepp, line 124: **Blocker: `issueKey` is used as a raw path segment — path traversal possible**

`path.join` resolves `..` normally, so a caller passing `issueKey = '../../evil'` produces a `targetDir` that escapes the session cache entirely:

```
path.join('/tmp/jira-mcp/session', 'issues', '../../evil', 'attachments')
// => /tmp/jira-mcp/evil/attachments   (outside session dir)
```

Even though real Jira keys follow `PROJECT-123`, the function accepts any string. When PR #7/#8 wire this up to a tool, the tool layer's input validation is the only guard — a missing or weak check there becomes a write-primitive to arbitrary subdirectories under `/tmp`.

Suggested fix in `attachmentDir` (or at call-site before passing to it):

```ts
function attachmentDir(issueKey: string): string {
  // Validate the key matches Jira's own format before using it as a path segment.
  if (!/^[A-Z][A-Z0-9_]+-[0-9]+$/i.test(issueKey)) {
    throw new Error(`Invalid issue key: ${issueKey}`);
  }
  return path.join(sessionCacheDir(), 'issues', issueKey, 'attachments');
}
```

Alternatively, apply the same basename-only treatment used for filenames.

src/core/attachments.ts
```diff
@@ -0,0 +1,231 @@
...
+    const errText = await res.bodyText().catch(() => "");
+    throw new Error(
+      `Failed to download attachment ${input.id} (${filename}): HTTP ${res.statusCode} ${errText.slice(0, 200)}`,
+    );
+  }
+
+  // Stream to disk. Writing to a temp file and renaming keeps the
+  // idempotency check honest: a partial file from an interrupted
+  // download won't pass the size check on the next call.
+  const tempPath = `${targetPath}.${process.pid}.partial`;
```
scottlepp: **Blocker: temp file uses only `process.pid` — concurrent downloads of the same attachment corrupt each other**

`process.pid` is constant for the lifetime of the process. Two concurrent `downloadAttachment` calls for the same `(issueKey, filename)` pair will both compute the same `tempPath` and both `pipeline` into it simultaneously:

```ts
const tempPath = `${targetPath}.${process.pid}.partial`;
// Both callers get: /…/notes.txt.12345.partial
// Both open WriteStreams to the same file and interleave writes.
```

MCP servers receive concurrent tool calls. If PR #7/#8 expose an attachment tool, two parallel agent steps requesting the same attachment will race here. The result is a corrupt file that passes the subsequent size check (or fails it and is deleted, losing both downloads).

Fix: add enough entropy to make the name unique per call:

```ts
import { randomBytes } from 'node:crypto';
const tempPath = `${targetPath}.${process.pid}.${randomBytes(6).toString('hex')}.partial`;
```

src/core/attachments.ts
```diff
@@ -0,0 +1,231 @@
...
+const TEXT_MIME_PATTERN = /^(text\/|application\/(json|xml|x-yaml|yaml|javascript))/i;
+
+async function extractPreview(
+  filePath: string,
+  mimeType: string,
+  maxChars: number,
+): Promise<string | null> {
+  if (!TEXT_MIME_PATTERN.test(mimeType)) return null;
+  try {
+    const buf = await fs.readFile(filePath);
```
scottlepp: **Suggestion: `extractPreview` reads the entire file into memory, defeating the no-heap-buffering goal for large text files**

The PR description emphasises that a 50 MB attachment never hits the Node heap during streaming download. But `extractPreview` immediately negates that for text/JSON files:

```ts
const buf = await fs.readFile(filePath);   // allocates entire file as Buffer
const text = buf.toString('utf8');         // second full copy as string
return text.slice(0, maxChars);            // returns ~2000 chars; rest is GC'd
```

A 50 MB `.json` or `.log` attachment allocates ~100 MB here (Buffer + string) then throws 99.9% away. Use a streaming read limited to `maxChars` bytes instead:

```ts
import { open } from 'node:fs/promises';

async function extractPreview(filePath: string, mimeType: string, maxChars: number) {
  if (!TEXT_MIME_PATTERN.test(mimeType)) return null;
  try {
    const fh = await open(filePath, 'r');
    try {
      const buf = Buffer.allocUnsafe(maxChars * 4); // UTF-8 worst case 4 bytes/char
      const { bytesRead } = await fh.read(buf, 0, buf.length, 0);
      const text = buf.subarray(0, bytesRead).toString('utf8');
      // Find a clean char boundary and check if more data follows.
      const full = bytesRead < buf.length;
      return full ? text : `${text.slice(0, maxChars)}…`;
    } finally {
      await fh.close();
    }
  } catch {
    return null;
  }
}
```

src/core/attachments.ts
```diff
@@ -0,0 +1,231 @@
...
+
+// --- Default transport -------------------------------------------------
+
+const defaultTransport: AttachmentTransport = async (url, init) => {
+  const res = await undiciRequest(url, {
+    method: init.method,
+    headers: init.headers,
+  });
+  return {
+    statusCode: res.statusCode,
```
scottlepp: **Suggestion: `defaultTransport` has a dual-read hazard between `body` and `bodyText`**

The returned object wraps the same undici `BodyReadable` in two different ways:

```ts
body: Readable.from(res.body),     // lazy wrapper — starts pulling when piped
bodyText: () => res.body.text(),   // calls undici directly — also pulls from res.body
```

Today this is safe because `bodyText` is only called in the error branch (where `body` has never been read from) and the success branch never calls `bodyText`. But the `AttachmentTransport` contract doesn't enforce this invariant — any future caller that reads `body` partially and then falls through to call `bodyText` (e.g., an error detected after some bytes are piped) will get an empty or truncated error message, and undici may not drain correctly.

Consider closing the hazard in `defaultTransport` itself by draining `body` through a single path, or add a guard that makes calling both an explicit error:

```ts
let bodyConsumed = false;
return {
  body: (() => { bodyConsumed = true; return Readable.from(res.body); })(),
  bodyText: async () => {
    if (bodyConsumed) throw new Error('body already consumed as stream');
    return res.body.text();
  },
};
```

scottlepp:
2 blockers, 2 suggestions found. The overall structure (streaming pipeline, atomic rename, idempotency, filename sanitisation) is solid — the issues are in the seams.

**Blockers**

- `src/core/attachments.ts:109` — `issueKey` is passed raw to `path.join` as a directory segment. `path.join` resolves `..` components normally, so a caller supplying `../../evil` escapes the session cache entirely. No sanitisation or format validation exists. The filename code does this correctly; the same treatment needs to apply to the key.

- `src/core/attachments.ts:194` — `tempPath` is derived purely from `process.pid`, which is constant for the process lifetime. Two concurrent `downloadAttachment` calls for the same `(issueKey, filename)` pair produce the same `tempPath` and both pipe into it simultaneously, corrupting the data. MCP servers field concurrent tool calls, so this is a realistic race. Adding `crypto.randomBytes(6).toString('hex')` to the suffix is sufficient.

**Suggestions**

- `src/core/attachments.ts:139` — `extractPreview` calls `fs.readFile` (full file into a Buffer, then `.toString()` into a second full string) before slicing to `maxChars`. For a 50 MB text/JSON attachment this allocates ~100 MB of heap and throws away 99.9% of it, directly contradicting the PR's stated design goal. Reading only `maxChars * 4` bytes via a file-handle `fh.read()` call eliminates the buffering.

- `src/core/attachments.ts:120` — `defaultTransport` exposes the same undici `BodyReadable` through both `body` (as a `Readable.from` wrapper) and `bodyText` (direct `.text()` call). The current call-sites are safe, but the `AttachmentTransport` type doesn't encode the mutual-exclusion invariant, making it a latent bug for future callers. A simple consumed-flag guard or a comment in the interface would close the hazard.

src/core/attachments.ts
scottlepp: Fixed in d057b3d. Added `assertValidIssueKey()` which requires the canonical Jira key shape `/^[A-Z][A-Z0-9_]*-[0-9]+$/` and is called before any fs work (so a bad key throws before mkdir even runs). Tests cover traversal (`../evil`), lowercase, missing parts, leading digits, slashes, empty — 14 rejection cases in total. The export is also available for tools layers to pre-validate.

src/core/attachments.ts
scottlepp: Fixed in d057b3d. Temp path now includes `randomBytes(6).toString('hex')` so concurrent calls get distinct temp files. One thing worth calling out that you only touched on briefly: after both temp files succeed, they both `fs.rename` onto the same target. Since the bytes are the same (same attachment), whichever renames last wins and the result is correct either way. A new test runs two parallel downloads of the same attachment and asserts the final file is uncorrupted.

src/core/attachments.ts
scottlepp: Fixed in d057b3d. `extractPreview` now opens the file and reads at most `maxChars * 4 + 4` bytes via `fh.read` rather than `fs.readFile` on the whole thing. The `+ 4` sentinel byte lets us distinguish "file ends within the limit" from "file extends beyond the limit, truncated" so the ellipsis is only appended when there's actually more content. New test passes a 5MB text attachment and verifies the preview still caps at the configured length.

Good catch — you're right that I crowed about streaming the download and then turned around and buffered the file for preview.

src/core/attachments.ts
scottlepp: Fixed in d057b3d. `defaultTransport` now tracks a `consumed: 'none' | 'stream' | 'text'` flag and throws `body already consumed via bodyText` (or vice versa) when both are touched on the same response. Zero overhead on the happy path — just a property access check. New unit test verifies the guard fires when a stream is consumed and then `bodyText()` is called.

Agreed that today's caller is safe. Future-proofing this was worth the 6 extra lines.

src/core/attachments.ts
```diff
@@ -0,0 +1,289 @@
...
+  // The response body is a single-consumption stream. Callers that
+  // read it as `body` (for streaming to disk) must not then also call
+  // `bodyText`, and vice versa. Today's callers respect this (success
+  // path only reads `body`, error path only reads `bodyText`), but
+  // this guard makes accidental double-consumption fail loudly rather
+  // than quietly returning garbage if a future caller forgets.
+  let consumed: "none" | "stream" | "text" = "none";
+  return {
+    statusCode: res.statusCode,
+    get body(): Readable {
```
scottlepp, line 149: **Suggestion: `body` getter does not guard against being called twice**

The guard checks `consumed === 'text'` but not `consumed === 'stream'`. If the getter is accessed a second time after the first call has already set `consumed = 'stream'`, it silently wraps the same undici body iterator in a second `Readable.from(res.body)`. The second consumer steals bytes from the first (or gets an empty stream), corrupting the download without any error.

The fix is a one-liner: add a `consumed === 'stream'` check symmetric to the `consumed === 'text'` check:

```ts
get body(): Readable {
  if (consumed === 'text') throw new Error('Response body already consumed via bodyText()');
  if (consumed === 'stream') throw new Error('Response body already consumed via body getter');
  consumed = 'stream';
  return Readable.from(res.body);
},
```

Current callers only read `.body` once (line 254), so there's no live bug today — but the guard was added precisely to catch this class of mistake and it doesn't cover this case.

scottlepp:
All 4 prior findings were fixed. One new suggestion found.

**Prior findings status:**

1. Blocker (issueKey path traversal): **Fixed.** `assertValidIssueKey` enforces `/^[A-Z][A-Z0-9_]*-[0-9]+$/` before any `path.join` call. Test coverage added.
2. Blocker (temp file pid-only collision): **Fixed.** Temp path now appends `randomBytes(6).toString('hex')`, giving ~281 trillion distinct suffixes per pid. Concurrent collision is no longer possible.
3. Suggestion (extractPreview full-file read): **Fixed.** `fh.read(buf, 0, readBytes, 0)` with `readBytes = maxChars * 4 + 4` caps the read at the byte level. Test confirms a 5 MB file produces a 501-char capped preview.
4. Suggestion (body/bodyText mutual exclusion not type-enforced): **Fixed.** The `consumed` state flag now makes both directions fail loudly. A unit test for the guard is included.

**New finding:**

- Suggestion — `src/core/attachments.ts:143` — The `body` getter guards against `consumed === 'text'` but not against `consumed === 'stream'`. Calling `.body` a second time silently produces a second `Readable.from(res.body)` wrapping the same undici iterator, which would steal or miss bytes without any thrown error. No live bug today (current callers read `.body` exactly once), but the guard is supposed to prevent this class of mistake. See inline comment for the one-line fix.

src/core/attachments.ts
scottlepp: Fixed in 71c3fff. Good catch — the guard was asymmetric. Same flaw existed on `bodyText` (double-text wasn't blocked either), so I extended to cover both.

Refactored into an exported `guardSingleConsumption()` helper so tests can exercise the real production code path rather than reimplementing the guard against a mock (which was how my first round's test slipped past this case). 5 new tests cover all 4 illegal transitions (stream→text, text→stream, stream→stream, text→text) plus both single-read happy paths, and assert the underlying consumers aren't called when the guard rejects.

scottlepp:
Round 3 review — no new issues found.

**Round-2 finding (asymmetric stream-guard): fixed.** `guardSingleConsumption` now symmetrically blocks all four bad-consumption orders (stream→bodyText, bodyText→stream, stream→stream, bodyText→bodyText) with dedicated error messages. All six cases are exercised in the new `guardSingleConsumption` test suite.

Reviewed the current diff for fresh issues:

- `extractPreview` truncation logic: the `truncatedRead && text.length > maxChars` invariant holds because `readBytes = maxChars * 4 + 4`; even at 4 bytes/char that yields `maxChars + 1` chars minimum, so `truncatedRead === true` always implies the file has more content past `maxChars`. No false ellipsis.
- Concurrent rename race: both callers download the same bytes to distinct temp files; the last `rename` wins atomically. Both subsequent `stat` checks pass since the winner's content is valid.
- `defaultTransport` encapsulates the raw undici `BodyReadable`; no caller outside the guard wrapper can touch it directly.
- `sanitizeFilename` handles null bytes and all path-separator variants. Remaining control chars (`\x01`–`\x08` etc.) are benign on Unix — not a blocker.
- Zero-size attachments are handled correctly by both the download and idempotency paths.

No blockers, no suggestions. PR is clean.

PR #172, opened 2026-04-24
v2: central operation manifest (PR #6)

scottlepp:
## Summary

Lands the single declaration every later PR consumes. Nothing is wired to the MCP server yet — this is pure plumbing + tests. Both Layer 2 (classic tools, PR #7) and Layer 3 (code-api generator, PR #8) will read from this manifest.

### \`src/core/manifest.ts\`
- **Types**: \`Operation\`, \`Manifest\`, \`ParamSpec\`, \`ParamRole\`, \`HttpVerb\`
- **Path helpers**: \`extractPathParams\` (for the generator), \`interpolatePath\` (URI-encodes values, throws on missing)
- **\`splitArgs\`**: partitions a flat args bag into path/query/body buckets, reports unknowns and missing-required separately so the caller can decide the policy
- **\`invokeOperation\`**: the single dispatcher; routes to \`JiraClient.get/post/put/delete\` or \`agile*\` variants based on verb + \`isAgile\`. Applies trim projections when specified.
- **\`OperationError\`**: structured error for unknown-op / missing-param cases

### \`src/core/trim-registry.ts\`
Named map of projections keyed by \`TrimKey\` (\`"issue" | "search" | "comment" | "attachment" | "user" | "project"\`). Lets the manifest reference projections as strings rather than holding function references — important because the Layer 3 generator will emit trim hints into stubs without needing to serialize closures.

### \`src/core/operations.ts\`
10 initial operations covering every shape the dispatcher/generator must handle:
| Shape | Op |
|---|---|
| GET + path + query | \`issue.get\` |
| POST + body | \`issue.create\` |
| PUT + path + body | \`issue.update\` |
| DELETE + path | \`issue.delete\` |
| JQL search (POST + body) | \`search.issues\` |
| GET nested collection | \`comment.list\` |
| POST nested resource | \`comment.add\` |
| Metadata GET (LRU-cacheable) | \`field.list\` |
| Agile GET | \`board.get\` |
| Agile GET + query | \`sprint.listForBoard\` |

PR #7 will fill in the remaining ~75 operations when building the consolidated classic tools.

## Design choices worth flagging

- **Stringly-typed args**: the manifest uses \`Record<string, unknown>\` at the dispatcher boundary. Type-safe wrappers are the generator's job — keeping the manifest shape small means we can iterate it generically without type gymnastics.
- **URI encoding in path interpolation**: values like \`a/b c\` become \`a%2Fb%20c\`. Tested.
- **Array query params join with commas** (\`fields=summary,status\`). That's Jira's convention across both platform and agile APIs.
- **Trim via registry, not closures**: keeps the manifest declaration serializable.

## Test plan

- [x] 20 new tests cover every helper + the dispatcher across all verb/agile/trim combinations
- [x] 167 / 167 core tests pass across 5 consecutive runs
- [x] \`npm run build\` clean
- [ ] Reviewer: is the \`TrimKey\` registry the right coupling? Alternative was to allow inline trim functions in \`Operation\`, but that would prevent serializing the manifest for the generator.

## Next up

- **PR #7**: consolidated Layer 2 tools (\`jira_issue\`, \`jira_search\`, … 15 total) that dispatch through this manifest.

### DIFF
--- a/src/core/manifest.ts
+++ b/src/core/manifest.ts
@@ -112,7 +112,12 @@ export function splitArgs(
   for (const spec of op.params) {
     known.add(spec.name);
     const raw = args[spec.name];
-    if (raw === undefined) {
+    // Treat explicit null the same as undefined: it can't satisfy a
+    // required param, and it'd be wrong to forward as a path segment
+    // or JSON body value. Falling through would produce a plain Error
+    // from interpolatePath instead of the OperationError callers
+    // expect.
+    if (raw === undefined || raw === null) {
       if (spec.required) missingRequired.push(spec.name);
       continue;
     }
--- a/src/core/manifest.ts
+++ b/src/core/manifest.ts
@@ -1,0 +1,234 @@
+// Central operation manifest.
+//
+// Single declaration of every Jira operation this server exposes. Two
+// consumers read this at runtime:
+//
+//   - Layer 2 (classic MCP tools, PR #7): a consolidated tool like
+//     `jira_issue` takes `{ action, ...args }` and dispatches through
+//     `invokeOperation` against the manifest entry whose name matches.
+//
+//   - Layer 3 (code-api stubs, PR #8): on server startup we emit one
+//     TypeScript file per operation under `${sessionCacheDir}/api/…`.
+//     Each stub is ~20 lines — a typed signature + a thin body that
+//     forwards to the running server via the local IPC bridge. The
+//     generator walks the manifest to produce these.
+//
+// The manifest is deliberately stringly-typed: operations declare
+// their params by name + role (path/query/body), not by a static
+// TypeScript shape. Type-safe wrappers are the *generator's* job in
+// Layer 3. Keeping the manifest shape small means we can iterate it
+// generically without type gymnastics.
+
+import type { JiraClient } from "../auth/jira-client.js";
+import { trimRegistry, type TrimKey } from "./trim-registry.js";
+
+// --- Types ------------------------------------------------------------
+
+export type HttpVerb = "GET" | "POST" | "PUT" | "DELETE";
+
+export type ParamRole = "path" | "query" | "body";
+
+export interface ParamSpec {
+  name: string;
+  role: ParamRole;
+  required?: boolean;
+  description?: string;
+}
+
+export interface Operation {
+  // Stable identifier — stable across v2 minor versions. Layer 3
+  // stubs are named after this (`issue.get` → `api/issues/getIssue.ts`).
+  name: string;
+  // Human-readable single-line summary. Surfaces in generated stubs'
+  // JSDoc and in Layer 2 tool schemas.
+  description: string;
+  verb: HttpVerb;
+  // Path template with `{paramName}` placeholders. Every placeholder
+  // must appear in `params` with `role: "path"`.
+  pathTemplate: string;
+  // Agile API vs Platform API — affects base URL selection. Mirrors
+  // the existing `isAgile` boolean in JiraClient.
+  isAgile?: boolean;
+  params: ParamSpec[];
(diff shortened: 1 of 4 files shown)


src/core/manifest.ts
```diff
@@ -0,0 +1,234 @@
...
+  const queryParams: Record<string, unknown> = {};
+  const body: Record<string, unknown> = {};
+  const known = new Set<string>();
+  const missingRequired: string[] = [];
+  let hasBody = false;
+
+  for (const spec of op.params) {
+    known.add(spec.name);
+    const raw = args[spec.name];
+    if (raw === undefined) {
```
scottlepp: **Bug: `null` bypasses the `missingRequired` check, causing a plain `Error` instead of `OperationError`**

`splitArgs` treats only `undefined` as absent (line 115: `if (raw === undefined)`). If a caller passes `null` for a required path param, `null` is placed into `pathParams`, the `missingRequired` list stays empty, and `invokeOperation` proceeds to `interpolatePath`, which then throws a plain `Error("Missing required path parameter: ...")" — not an `OperationError`.

Any caller that catches `OperationError` to distinguish "caller mistake" from "upstream failure" will silently miss this case.

Fix: treat `null` the same as `undefined` in the check:
```ts
if (raw === undefined || raw === null) {
  if (spec.required) missingRequired.push(spec.name);
  continue;
}
```
The existing test at line 96 only exercises `interpolatePath` directly; a test through `invokeOperation` (passing `null` for a required param and asserting the rejection is an `OperationError`) would cover this path.

scottlepp:
**Review summary — 1 blocker, 0 other findings**

The PR is clean overall. One correctness bug:

**Blocker — `src/core/manifest.ts` line 115**: `splitArgs` checks only `raw === undefined` to detect absent params. A caller passing `null` for a required path param (valid in `Record<string, unknown>`) slips past the `missingRequired` guard, lands in `pathParams`, and then causes `interpolatePath` to throw a plain `Error` — not an `OperationError`. Any Layer 2 or Layer 3 code that catches `OperationError` to route "bad caller input" errors will miss this case entirely.

Fix: change the guard to `raw === undefined || raw === null`. A complementary test through `invokeOperation` (required param set to `null`, asserting `rejects.toBeInstanceOf(OperationError)`) would close the coverage gap — the existing test at line 96 only exercises `interpolatePath` in isolation.

Everything else looks good: the type/interface design is clean, the switch arms exhaustively cover `HttpVerb`, the trim-registry indirection is the right call for serializability, and the 20 test cases cover the key dispatcher paths well.

src/core/manifest.ts
scottlepp: Fixed in ad63698. `splitArgs` now treats `null` and `undefined` identically for the required-param check, so null bubbles up as an `OperationError` instead of a plain `Error` from `interpolatePath`. Added two regression tests: the unit-level one on `splitArgs` and the integration-level one through `invokeOperation` that asserts `rejects.toBeInstanceOf(OperationError)` — exactly the assertion that would have caught this.

Good catch on the caller-handling implication. The `instanceof OperationError` branch is the whole reason the custom error class exists, and plain `Error` leaking through silently defeated that.

scottlepp:
Round 2 review — no new issues found.

**Prior blocker (splitArgs null guard) — fixed.** The guard at manifest.ts line 120 now reads `raw === undefined || raw === null`, so null-valued required params are caught in `splitArgs` and surface as `OperationError` rather than leaking to `interpolatePath` as a plain `Error`. Two regression tests were added (unit-level on `splitArgs`, integration-level through `invokeOperation`) confirming the fix.

No new blockers or suggestions from the updated diff. The dispatcher verb routing, query-param normalisation, trim-registry lookup, and test coverage all look correct.

PR #173, opened 2026-04-24
v2: complete operation manifest (PR #7a)

scottlepp:
## Summary

First of three sub-PRs for what the plan listed as PR #7. Expands the manifest from the 10-op sample landed in PR #6 to the complete 85-op manifest covering every Jira API call v1 makes. **Purely declarative — no runtime wiring.** PRs #7b and #7c will consume it to build the consolidated classic tool handlers.

## Why three sub-PRs?

The original PR #7 would have been ~2500 lines (75 new manifest entries + 15 consolidated handlers + routing + Zod schemas + tests). Splitting into three:

- **7a (this PR):** manifest declarations + shape invariant tests
- **7b:** the generic dispatcher wrapper + first 3-4 consolidated tool handlers — proves the shape end-to-end
- **7c:** remaining 11 tool handlers + v1 tool file deletion

Each sub-PR is reviewable independently and builds on the last.

### \`src/core/operations.ts\`

75 new \`Operation\` entries on top of the 10 from PR #6, grouped by category. Every entry derived by reading v1's \`client.get/post/put/delete\` call site so paths, verbs, and param roles mirror v1 verbatim.

| Category | Count | | Category | Count |
|---|---|---|---|---|
| issue | 9 | | sprint | 8 |
| search | 3 | | epic | 4 |
| comment | 4 | | worklog | 4 |
| attachment | 3 | | filter | 6 |
| project | 11 | | issueLink | 4 |
| user | 5 | | watcher | 3 |
| board | 8 | | vote | 3 |
| field | 6 | | group | 3 |
| server | 1 | | | |

### One correctness fix

\`search.issues\` now points at **GET /search/jql** (v1's actual endpoint), not POST /search (deprecated). The PR #6 sample had the wrong endpoint — this would have been a runtime failure the moment 7b wired up the search tool. The reviewer on PR #172 asked about this explicitly; acting on it now before it bites.

### Also noticed during survey

Added \`issueLink.get\` (\`GET /issueLink/{linkId}\`) — v1 exposes this but I missed it in the initial sample.

### Intentionally left out

\`jira_get_attachment_content\` — v1 returned a download URL that agents couldn't actually use because they weren't authenticated. Replaced by the streaming downloader in \`src/core/attachments.ts\` from PR #5, which isn't a REST call and so has no manifest entry.

## Tests

\`tests/core/operations.test.ts\` (93 tests):

- **Shape invariants**: unique names; every \`{foo}\` placeholder has a matching \`role: "path", required: true\` param (and no orphan path params); no duplicate param names within an op; only supported verbs; body params only on POST/PUT; non-empty descriptions for every op.
- **Category coverage**: every v1 category's critical ops are declared. If a future PR accidentally drops one, this test fails.
- **Agile routing**: any path under \`/board\`, \`/sprint\`, \`/epic\`, \`/backlog\` must flag \`isAgile: true\` (and nothing else should). Catches the whole \"why is my board call hitting the platform URL\" class of bug.

## Test plan

- [x] 93 new tests in operations.test.ts; 262/262 core tests pass across 3 consecutive runs
- [x] \`npm run build\` clean
- [ ] Reviewer: spot-check a few manifest entries against [Jira REST docs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/) for endpoints you're familiar with. The survey was mechanical but the source of truth is Atlassian, not v1.
- [ ] Reviewer: worth raising a bug for any v1 operation you think should have been omitted (legacy, deprecated, overlaps with a newer endpoint)?

## Next up

- **PR #7b**: generic consolidated tool handler + \`jira_issue\`, \`jira_search\`, \`jira_comment\`, \`jira_user\` wired through it

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -752,7 +752,10 @@ export const operations: Manifest = [
       { name: "filterName", role: "query" },
       { name: "accountId", role: "query" },
       { name: "owner", role: "query" },
-      { name: "groupname", role: "query" },
+      // Note: Jira deprecated `groupname` here in favour of
+      // `groupId` (Cloud REST v3). v2 is a clean break, so we go
+      // straight to the new param.
+      { name: "groupId", role: "query" },
       { name: "projectId", role: "query" },
       { name: "id", role: "query" },
       { name: "orderBy", role: "query" },
@@ -865,6 +868,11 @@ export const operations: Manifest = [
     description: "Add a watcher to an issue.",
     verb: "POST",
     pathTemplate: "/issue/{issueIdOrKey}/watchers",
+    // Jira's POST /issue/{key}/watchers expects a raw JSON string
+    // body (e.g. `"acc123"`), not a JSON object. Without bodyShape,
+    // the dispatcher would send `{"accountId":"acc123"}` which Jira
+    // rejects with HTTP 400.
+    bodyShape: "rawString",
     params: [
       { name: "issueIdOrKey", role: "path", required: true },
       { name: "accountId", role: "body", required: true },
@@ -975,12 +983,16 @@ export const operations: Manifest = [
     verb: "GET",
     pathTemplate: "/group/member",
     params: [
-      { name: "groupname", role: "query", required: true },
+      { name: "groupId", role: "query", required: true },
       { name: "includeInactiveUsers", role: "query" },
       { name: "startAt", role: "query" },
       { name: "maxResults", role: "query" },
     ],
   },
+
+  // =================================================================
+  // Permissions
+  // =================================================================
   {
     name: "permissions.mine",
     description: "List permissions the current user has.",
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -1,25 +1,26 @@
-// Initial operation declarations.
+// Complete operation manifest.
 //
-// This PR (#6) lands a representative subset covering every shape the
-// manifest / dispatcher / generator must handle:
-//   - GET with path + query params                     (issue.get)
-//   - GET returning a paginated collection             (comment.list)
-//   - POST with JSON body                              (issue.create)
-//   - POST with path + body (nested resource)          (comment.add)
-//   - PUT with path + body                             (issue.update)
-//   - DELETE with path                                 (issue.delete)
-//   - JQL search (POST with body, not GET)             (search.issues)
-//   - Agile API with path params                       (board.get)
-//   - Agile API with path + paginated query            (sprint.listForBoard)
-//   - Metadata GET (cached by JiraClient's LRU)        (field.list)
+// One Operation per discrete Jira API call the server supports. Both
+// Layer 2 (consolidated classic tools, PR #7b/#7c) and Layer 3
+// (code-api stubs, PR #8) read this at runtime; no Jira call should
+// be made outside the dispatcher in `manifest.ts`.
 //
-// PR #7 extends this to cover the full v1 surface (~85 operations)
-// before the consolidated classic tools go live.
+// Naming convention: `category.verb` where `verb` reflects the
+// logical action (not the HTTP verb). Plural categories for
+// collection-level ops (`sprints.listForBoard`), singular for
+// entity ops (`sprint.get`).
+//
+// Derivation: each entry was translated from the corresponding v1
+// tool handler by reading its `client.get/post/put/delete` (or
+// agile* variant) call site. Paths, verbs, and param roles are
+// taken verbatim from v1 to preserve behavior.
 
 import type { Manifest } from "./manifest.js";
 
 export const operations: Manifest = [
-  // ------------------------------- Issue --------------------------------
+  // =================================================================
+  // Issue
+  // =================================================================
   {
     name: "issue.get",
     description: "Fetch a single issue by key or id.",
@@ -27,8 +28,8 @@ export const operations: Manifest = [
     pathTemplate: "/issue/{issueIdOrKey}",
     params: [
       { name: "issueIdOrKey", role: "path", required: true },
-      { name: "fields", role: "query", description: "Comma-separated field list" },
-      { name: "expand", role: "query", description: "Comma-separated expand list" },
+      { name: "fields", role: "query" },
+      { name: "expand", role: "query" },
     ],
     trim: "issue",
   },
@@ -47,7 +48,7 @@ export const operations: Manifest = [
   },
   {
     name: "issue.update",
-    description: "Update fields on an existing issue.",
+    description: "Update an existing issue.",
     verb: "PUT",
     pathTemplate: "/issue/{issueIdOrKey}",
     params: [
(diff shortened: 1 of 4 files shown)


src/core/operations.ts
```diff
@@ -139,7 +542,467 @@ export const operations: Manifest = [
...
+    params: [{ name: "issueIdOrKey", role: "path", required: true }],
+  },
+  {
+    name: "watcher.add",
+    description: "Add a watcher to an issue.",
+    verb: "POST",
+    pathTemplate: "/issue/{issueIdOrKey}/watchers",
+    params: [
+      { name: "issueIdOrKey", role: "path", required: true },
+      { name: "accountId", role: "body", required: true },
```
scottlepp, line 878: **Blocker: body shape mismatch — will send `{"accountId":"..."}` instead of a raw JSON string.**

The Jira REST API for `POST /issue/{issueIdOrKey}/watchers` requires the request body to be a bare JSON string (`"acc123"`), not a JSON object. The dispatcher's `splitArgs` always assembles body params into a `Record<string, unknown>`, so this param will be serialized as `{"accountId":"acc123"}`, which Jira will reject with a 400.

V1's `watchers.ts` (line 165) handles this correctly by calling `JSON.stringify(accountId)` to produce the raw string.

The manifest has no first-class concept of a "raw string body" — the body is always an object. Options for PR #7b:
- Add a special `role: "rawBody"` param role that the dispatcher forwards verbatim instead of wrapping.
- Or handle `watcher.add` as a one-off in the Layer 2 tool handler rather than through the generic dispatcher.

src/core/operations.ts
```diff
@@ -139,7 +542,467 @@ export const operations: Manifest = [
...
+    pathTemplate: "/group/member",
+    params: [
+      { name: "groupname", role: "query", required: true },
+      { name: "includeInactiveUsers", role: "query" },
+      { name: "startAt", role: "query" },
+      { name: "maxResults", role: "query" },
     ],
   },
+  {
+    name: "permissions.mine",
```
scottlepp, line 997: **Suggestion: `permissions.mine` name prefix doesn't match its category bucket.**

This operation's name starts with `permissions.` but it lives in the `group` section of the file and is tested under the `"group"` key in the category coverage table (test line 168). Any tooling that derives category by splitting on `.` (e.g. a Layer 3 generator grouping stubs into directories) will route it to a `permissions/` directory instead of `group/`. This disagreement between the name and the grouping will become a concrete bug in PR #8.

Either rename it to `group.permissions` / `group.myPermissions`, or move it to its own top-level `permissions` section (with a matching entry in the test's `required` map).

src/core/operations.ts
```diff
@@ -139,7 +542,467 @@ export const operations: Manifest = [
...
+      { name: "maxResults", role: "query" },
+    ],
+  },
+  {
+    name: "group.members",
+    description: "List members of a group.",
+    verb: "GET",
+    pathTemplate: "/group/member",
+    params: [
+      { name: "groupname", role: "query", required: true },
```
scottlepp: **Suggestion: `groupname` is deprecated in Jira Cloud REST v3 — use `groupId` instead.**

The Jira Cloud platform REST API v3 deprecated the `groupname` query parameter for `GET /group/member` (and `GET /filter/search`) in favour of `groupId`. Passing `groupname` still works today but Atlassian marks it as deprecated and may remove it. Since the v2 rewrite is targeting Cloud v3 (evidenced by the `/search/jql` change), it's worth switching before PR #7b wires this up — changing a required param name later will be a breaking change to the Layer 2 tool schema.

scottlepp:
1 blocker, 2 suggestions across 3 line comments.

**Blocker — `src/core/operations.ts` line 870 (`watcher.add`):** The Jira API for `POST /issue/{issueIdOrKey}/watchers` takes a bare JSON string as the body (e.g. `"acc123"`), not an object. The dispatcher's `splitArgs` always wraps body-role params into a `Record<string, unknown>`, so this will send `{"accountId":"acc123"}` and Jira will reject it with a 400. V1's `watchers.ts:165` works around this by calling `JSON.stringify(accountId)` directly. The manifest's body model has no concept of a "raw string body" — needs a `role: "rawBody"` escape hatch or a one-off handler bypass before PR #7b can wire this up correctly.

**Suggestion — line 985 (`permissions.mine`):** Operation name prefix is `permissions.` but the operation is filed under the `group` section and under the `"group"` key in the test coverage table. Any Layer 3 generator that derives directory/stub grouping from the name prefix will put this stub in `permissions/` instead of `group/`. Should either rename to `group.myPermissions` or add a first-class `permissions` section and update the test accordingly.

**Suggestion — line 978 (`group.members`):** `groupname` is deprecated in Jira Cloud REST v3 in favour of `groupId`. Also affects `filter.list` line 755. Since this PR explicitly targets the v3 `/search/jql` endpoint, it's cheaper to update the param name now than after PR #7b exposes it in a tool schema.

src/core/operations.ts
scottlepp: Fixed in f643198. Went with a third option you didn't mention — added `Operation.bodyShape: "object" | "rawString"` (defaults to "object"). With `rawString`, the operation must declare exactly one body param and the dispatcher forwards its raw value, so the wire format becomes the literal JSON string `"acc123"` instead of `{"accountId":"acc123"}`.

Reasoning vs. the alternatives: a new `role: "rawBody"` would create ambiguity with regular `body` params, and per-handler escape hatches in PR #7b would scatter wire-format knowledge away from the manifest. `bodyShape` keeps the manifest self-describing.

Two new tests:
- Manifest-level invariant: any rawString op must declare exactly one body param. Catches misconfigured manifests at the test gate.
- Dispatcher-level: watcher.add with `{key, accountId}` sends `body: "acc123"` (literal string), not an object.

src/core/operations.ts
scottlepp: Fixed in f643198. Pulled `permissions.mine` into its own `// Permissions` section and added a top-level `permissions:` key in the test category-coverage map.

Also added a generic shape invariant that catches this whole class: every op name must match `<category>.<verb>` with a non-empty category. So if a future PR introduces another name/section divergence, the test catches it before it bites the generator.

src/core/operations.ts
scottlepp: Fixed in f643198. Renamed `groupname` → `groupId` in both `group.members` (where it was required) and `filter.list` (where it was optional). Added an inline note on `filter.list` so future readers know why we use the new param.

Agreed on the breaking-change risk — better to take the breakage now in the manifest than to bake `groupname` into Layer 2's Zod schema and have to migrate callers later.

scottlepp:
**Round 2 review — all three prior findings fixed; no new blockers.**

Prior findings status:

1. **Blocker (operations.ts watcher.add — raw string body)** — Fixed. `BodyShape = "object" | "rawString"` added to manifest; `invokeOperation` extracts the scalar value and forwards it unwrapped. Two new tests cover the happy path and the invalid-arity invariant.

2. **Suggestion (permissions.mine misrouted under group)** — Fixed. `permissions.mine` now lives in its own `// Permissions` section with a matching `permissions:` key in the category-coverage test. A new shape invariant ("name must be `<category>.<verb>` with non-empty category") catches any future recurrence.

3. **Suggestion (groupname deprecated, prefer groupId)** — Fixed. Both `group.members` and `filter.list` now use `groupId`. An inline comment on `filter.list` records why.

New diff (manifest.ts bodyShape + operations.ts 75 new entries + operations.test.ts 93 new tests):

No new blockers. One nit worth noting but not a blocker: the `rawString` fallback at manifest.ts line 232-235 returns `undefined` when the key is absent from `split.body`. This is dead code in practice because `required: true` on `watcher.add`'s `accountId` param causes `invokeOperation` to throw before reaching that branch — but it silently swallows a misconfiguration where `bodyShape: "rawString"` is combined with `required: false`. Not a runtime risk today; fine to leave or add a comment.

Overall the manifest is well-structured, the invariant tests are comprehensive, and the `rawString` fix correctly addresses the Jira wire-format constraint.

PR #174, opened 2026-04-27
v2: generic dispatcher + first 4 consolidated tools (PR #7b)

scottlepp:
## Summary

Second of three sub-PRs for what the plan listed as PR #7. Lands the runtime layer that turns the manifest from PR #7a into working MCP tools. **Behind \`JIRA_V2_TOOLS=1\` (off by default)** — v1 surface is untouched. PR #7c will add the remaining 11 tools and remove v1.

### \`src/tools/v2/dispatcher.ts\`
Shared dispatcher every consolidated tool uses. Steps:

1. Pull \`action\` off args.
2. Look up the action's manifest operation + Zod schema.
3. Validate (ZodError → ToolError with field-by-field message).
4. Forward to \`invokeOperation\` (OperationError → ToolError).

Also exports \`buildInputSchema()\` which emits the JSON Schema served on \`tools/list\`. Uses **\`oneOf\` with a \`const\` action discriminator per branch** so agents see a strict per-action input shape rather than a flat union.

### Hand-written Zod → JSON Schema converter
No \`zod-to-json-schema\` dep. Reasons:
- Zod 4 dropped \`_def.typeName\` / \`_def.shape()\`; existing converters target Zod 3 and would force a downgrade or wait for upstream.
- The surface I cover is small (string/number/boolean/array/object/record + optional/default/nullable wrapping). The converter is ~50 lines and explicit.
- One less transitive dep.

### Four consolidated tools

| v2 tool | Replaces | Actions |
|---|---|---|
| \`jira_issue\` | 9 v1 tools | get, create, update, delete, bulkCreate, listTransitions, transition, assign, changelog |
| \`jira_search\` | 2 v1 tools | issues, jqlAutocompleteData, jqlSuggestions |
| \`jira_comment\` | 4 v1 tools | list, add, update, delete |
| \`jira_user\` | 5 v1 tools | myself, search, get, assignable, bulkGet |

23 → 4 tools, 31 actions.

### Opt-in wiring
\`src/index.ts\` reads \`JIRA_V2_TOOLS\` once at startup. When \`=1\`:
- v2 tools are **appended** to the v1 listing (names don't overlap, so concat is safe — v1 is \`jira_<verb>_<noun>\`, v2 is \`jira_<category>\`).
- Calls to v2 names route through \`handleV2Tool\` first; v1 calls fall through unchanged.

When \`JIRA_V2_TOOLS\` is unset or anything other than \`1\`, behavior is identical to v1. No risk to existing users.

## Tests

**38 new tests, 296 / 296 unit tests pass across 3 consecutive runs.**

- \`tests/tools/v2/dispatcher.test.ts\` — dispatch (missing action, unknown action, validation failure, args strip, OperationError wrapping) + JSON schema generation (oneOf shape, optional vs required marking)
- \`tests/tools/v2/tools.test.ts\` — every action of every tool dispatches the right (verb, path, body, query) against the production manifest. **Includes a coverage test asserting every action's declared operation exists in the manifest** — guards typos as we add more tools in 7c.

## Test plan

- [x] \`npm run build\` clean
- [x] 296 / 296 core + tools tests pass × 3 runs
- [ ] Reviewer: do the per-tool action names feel right? (\`bulkCreate\` vs \`bulk\`, \`assign\` vs \`setAssignee\`, etc.)
- [ ] Reviewer: any concerns about \`oneOf\` schema rendering in agents you've used? Falling back to a flat \`anyOf\` or to a non-discriminated schema is straightforward if needed.

## Next up

- **PR #7c**: remaining 11 consolidated tools (project, board, sprint, epic, worklog, attachment, filter, link, watcher, vote, field, group, permissions, server) + v1 tool file deletion + flip the default to v2

### DIFF
--- a/src/tools/v2/dispatcher.ts
+++ b/src/tools/v2/dispatcher.ts
@@ -84,6 +84,8 @@ interface ZodDef {
   element?: ZodType;
   innerType?: ZodType;
   valueType?: ZodType;
+  // ZodUnion: array of constituent schemas.
+  options?: ZodType[];
 }
 
 function zodKind(schema: ZodType): string {
@@ -164,6 +166,15 @@ function zodFieldToJsonSchema(schema: ZodType): unknown {
     case "ZodRecord":
     case "ZodObject":
       return { type: "object", additionalProperties: true, ...tail };
+    case "ZodUnion": {
+      // Render each constituent as its own JSON Schema and combine
+      // via oneOf. Without this case, fields like
+      // z.union([z.string(), z.array(z.string())]) would emit only
+      // their description with no type info, leaving agents unable
+      // to satisfy the schema.
+      const opts = zodDef(inner).options ?? [];
+      return { oneOf: opts.map(zodFieldToJsonSchema), ...tail };
+    }
     case "ZodAny":
     case "ZodUnknown":
       return tail;
--- a/src/tools/v2/dispatcher.ts
+++ b/src/tools/v2/dispatcher.ts
@@ -1,0 +1,252 @@
+// Generic dispatcher for v2 consolidated tools.
+//
+// Every consolidated tool (jira_issue, jira_search, …) follows the
+// same shape:
+//   1. Args land here as `Record<string, unknown>`.
+//   2. We pull `action` off, look up the operation name via the
+//      tool's actionMap, then validate the rest with the action's
+//      Zod schema.
+//   3. The validated args go through `invokeOperation`, which
+//      handles path interpolation, body shaping, agile vs platform
+//      routing, and trim projections.
+//
+// All knowledge of *what* actions a tool exposes lives in the
+// `ConsolidatedTool` definition. The dispatcher is the same for
+// every tool.
+
+import { z, ZodError, type ZodType } from "zod";
+
+import type { JiraClient } from "../../auth/jira-client.js";
+import {
+  invokeOperation,
+  OperationError,
+  type Manifest,
+} from "../../core/manifest.js";
+
+// --- Tool definition shape --------------------------------------------
+
+export interface ActionDefinition {
+  // Human-readable single-line description of this action. Surfaces
+  // in the MCP tool's JSON schema so agents pick the right action.
+  description: string;
+  // Zod schema for the action's inputs (excluding `action` itself).
+  // Validated runtime; failures become OperationError-shaped responses.
+  schema: ZodType;
+  // The manifest operation invoked when this action is selected.
+  operation: string;
+}
+
+export interface ConsolidatedTool {
+  // The MCP tool name (e.g. "jira_issue").
+  name: string;
+  // Surfaces on the MCP tool listing.
+  description: string;
+  // Map from action key (e.g. "get") to its definition.
+  actions: Record<string, ActionDefinition>;
+}
+
+// --- MCP-shaped JSON schema for the tool listing ----------------------
+
+// Build the input schema the MCP tool listing exposes. Uses oneOf so
+// the agent's tool-use renderer constrains output per action.
+//
+// We don't go through zod-to-json-schema because (a) we want full
+// control over the wire shape, (b) it'd add a dep, and (c) the action
+// schemas are small enough that hand-shaped JSON is clearer.
+export function buildInputSchema(tool: ConsolidatedTool): unknown {
+  const oneOf: unknown[] = [];
+  for (const [actionName, def] of Object.entries(tool.actions)) {
+    oneOf.push({
+      type: "object",
+      title: actionName,
+      description: def.description,
+      properties: zodToProperties(def.schema, actionName),
+      required: zodRequired(def.schema, actionName),
+      additionalProperties: false,
+    });
+  }
+  return {
+    type: "object",
+    description: tool.description,
+    oneOf,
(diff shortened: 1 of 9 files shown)


src/tools/v2/dispatcher.ts
```diff
@@ -0,0 +1,252 @@
...
+        ...tail,
+      };
+    }
+    case "ZodRecord":
+    case "ZodObject":
+      return { type: "object", additionalProperties: true, ...tail };
+    case "ZodAny":
+    case "ZodUnknown":
+      return tail;
+    default:
```
scottlepp, line 181: **Bug (suggestion): `ZodUnion` silently falls through to `{}` in the schema converter**

`BulkGetSchema` in `user.ts` declares `accountId` as `z.union([z.string(), z.array(z.string())])`. The `switch` in `zodFieldToJsonSchema` has no `ZodUnion` case, so the constructor name `"ZodUnion"` hits this `default` and returns `{}` (or `{description}` if described). The tool listing exposes `accountId` as a completely unconstrained property — MCP clients and agents cannot tell a string or string-array is expected.

Runtime dispatch is unaffected (Zod validates correctly). The gap is only in the `inputSchema` served on `tools/list`.

**Fix:** add a `ZodUnion` case before the `default`:
```typescript
case "ZodUnion": {
  const options = (zodDef(inner) as { options?: ZodType[] }).options ?? [];
  return { oneOf: options.map(zodFieldToJsonSchema), ...tail };
}
```

There is also no `buildInputSchema` test covering a union-typed field; adding one would prevent future regression here.

src/tools/v2/user.ts
```diff
@@ -0,0 +1,57 @@
...
+  project: z.string().optional().describe("Project key — restricts to assignable for this project"),
+  issueKey: z.string().optional().describe("Issue key — restricts to assignable for this issue"),
+  startAt: z.number().optional(),
+  maxResults: z.number().optional(),
+  actionDescriptorId: z.number().optional(),
+  recommend: z.boolean().optional(),
+});
+
+const BulkGetSchema = z.object({
+  accountId: z.union([z.string(), z.array(z.string())]).describe("One or more accountIds"),
```
scottlepp, line 41: **Note:** Because `zodFieldToJsonSchema` in `dispatcher.ts` has no `ZodUnion` case, this field's JSON Schema is emitted as `{ description: "One or more accountIds" }` — no type information. See the comment on `dispatcher.ts` line 170 for the fix.

scottlepp:
**Review summary: 1 bug, 1 cross-reference note.**

**Blocker / suggestion — `ZodUnion` not handled in `zodFieldToJsonSchema` (`src/tools/v2/dispatcher.ts` line 170)**
The hand-rolled Zod-to-JSON-Schema converter covers `ZodString`, `ZodNumber`, `ZodBoolean`, `ZodArray`, `ZodRecord`, `ZodObject`, `ZodAny`, and `ZodUnknown`, but has no case for `ZodUnion`. The constructor name `"ZodUnion"` falls through to `default` and returns `{}` (or just a `{description}` object). The only current user is `BulkGetSchema.accountId` in `user.ts` (`z.union([z.string(), z.array(z.string())])`), which consequently gets emitted as `{ description: "One or more accountIds" }` on `tools/list`. Agents see no type constraint and cannot tell whether to pass a string or an array. The fix is one new `case "ZodUnion":` block — see the inline comment for the exact code. A test for `buildInputSchema` over a union-typed field would lock this in.

Everything else in the PR looks correct: the dispatch pipeline is sound, `ToolError` propagation is handled by the existing `instanceof Error` branch in `src/index.ts`, the manifest coverage test guards against typo-operation references, and the opt-in flag wiring is clean.

src/tools/v2/dispatcher.ts
scottlepp: Fixed in 4c74510. Added a `ZodUnion` case that maps each constituent through the converter and combines via `oneOf`. Verified the actual `jira_user.bulkGet` listing now exposes `accountId: { oneOf: [{ type: "string" }, { type: "array", items: { type: "string" } }], description }`.

Took the suggested fix shape almost verbatim. Added a regression test in `dispatcher.test.ts` that exercises a 3-way union (string + array + number) to lock in both the structure and the description-preservation behavior.

src/tools/v2/user.ts
scottlepp: Fixed via the dispatcher.ts change in 4c74510. `accountId` here now serializes as oneOf string vs string[] in the tool listing.

scottlepp:
Round 2 review — prior finding: **fixed**.

The `ZodUnion` case was added to `zodFieldToJsonSchema` in `dispatcher.ts` (lines 169-176). It maps each constituent through the converter and combines via `oneOf`, exactly as recommended. `BulkGetSchema.accountId` (`z.union([z.string(), z.array(z.string())])`) now emits a proper `oneOf` in the tool listing. A regression test covering union-typed fields was also added in `dispatcher.test.ts`.

No new issues found in the updated diff. The rest of the surface — `ZodDefault`/`ZodOptional`/`ZodNullable` unwrapping, action dispatch, `ToolError` propagation through the `index.ts` catch block, and the manifest-coverage guard test — all look correct.

PR #176, opened 2026-04-27
v2: code-api stub generator (PR #8)

scottlepp:
## Summary
- Walks the operation manifest and emits TypeScript stubs for every Jira operation under a target `api/` directory — the foundation of Layer 3 (code execution with MCP).
- Each stub is a typed args interface plus a thin function that forwards to a placeholder `_client.invoke`. The real IPC bridge lands in PR #9; until then `_client.ts` throws a clear error so misuse is loud rather than silent.
- Output is deterministic and idempotent: same manifest in → byte-identical files out, so the runtime can skip regeneration and so we can golden-test the result.

### What's new
- `src/codeapi/templates.ts` — pure render helpers (`renderStub`, `renderCategoryIndex`, `renderRootIndex`, `renderClientFile`, `renderTypesFile`). Splits dotted operation names (`issue.get` → `issue/get.ts`), sanitizes identifiers, sorts params required-first then by role (path/query/body) then alphabetically.
- `src/codeapi/generator.ts` — `planApi(manifest)` returns the file plan without touching disk; `generateApi({ manifest, outDir })` writes it. Groups operations by their category prefix.
- `tests/codeapi/generator.test.ts` — 18 tests: name splitting, stub structure, param ordering, agile/trim hints, idempotence, and an end-to-end run against the real 85-op manifest.

### Out of scope (for later PRs)
- IPC bridge that backs `_client.invoke` — PR #9.
- Wiring `generateApi()` into the MCP server's startup path / making `jira_code_api` the default tool — PR #10.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — all tests pass (existing 323 + 18 new = 341)
- [x] Smoke test: `generateApi({ manifest: operations, outDir })` produces 106 files for 85 operations across 18 categories; spot-checked stubs render cleanly with correct types and JSDoc


tests/codeapi/generator.test.ts
```diff
@@ -0,0 +1,322 @@
...
+
+// --- renderCategoryIndex ----------------------------------------------
+
+describe("renderCategoryIndex", () => {
+  it("re-exports each verb from a sibling file", () => {
+    const out = renderCategoryIndex("issue", [
+      fixtureManifest[0],
+      fixtureManifest[1],
+    ]);
+    expect(out).toContain('export { create } from "./create.js";');
```
scottlepp, line 188: The test suite has no case for a reserved-word verb. Nine operations in the live manifest use the verb `delete`, which `safeIdentifier` correctly renames to `delete_` in the stub. The side-effect is that callers using namespace import style get `issue.delete_()` instead of `issue.delete()`, which is surprising but intentional.

Please add at least one unit test here (or in the `renderCategoryIndex` block) that uses `attachment.delete` (or a synthetic op with verb `delete`) and asserts:
1. The stub file exports `function delete_` (not `function delete`).
2. The category index emits `export { delete_ } from "./delete.js";` — confirming the re-export name matches what the stub exports.

Without this, a refactor of `safeIdentifier` or the category-index template could silently break the only naming guarantee that prevents a JS syntax error in generated output.

src/codeapi/generator.ts
```diff
@@ -0,0 +1,156 @@
...
+  out.push({ relativePath: "_client.ts", contents: renderClientFile() });
+  out.push({
+    relativePath: "types.ts",
+    contents: renderTypesFile(refsImportPath),
+  });
+  out.push({
+    relativePath: "index.ts",
+    contents: renderRootIndex([...groups.keys()]),
+  });
+
```
scottlepp, line 103: The generator silently overwrites a file if two operations in the same category share the same verb (i.e., the same `category/verb.ts` path). `planApi` would still return `N` entries for `N` operations — so the `stubPaths.length === operations.length` assertion in the integration test would pass — but the last operation alphabetically by manifest order would win, and the first would be silently lost.

This cannot happen with today's 85-op manifest (verified), but `planApi` has no guard and the test suite won't catch it if a future manifest entry introduces a duplicate. Consider adding a collision check in `planApi`:

```ts
const seen = new Set<string>();
for (const file of out) {
  if (seen.has(file.relativePath)) {
    throw new Error(`planApi: duplicate output path: ${file.relativePath}`);
  }
  seen.add(file.relativePath);
}
```

A corresponding test with a manifest that has two operations mapping to the same path would lock this in.

src/codeapi/generator.ts
```diff
@@ -0,0 +1,156 @@
...
+export interface GenerateApiResult {
+  outDir: string;
+  files: GeneratedFile[];
+  operationCount: number;
+  categories: string[];
+}
+
+// Default refs import. Points at the build output relative to the
+// installed jira-mcp package. `jira-mcp/build/types/refs.js` is the
+// stable pin; the file already exists in the repo.
```
scottlepp: The default \`refsImportPath = "jira-mcp/build/types/refs.js"\` is a package-relative specifier that requires \`jira-mcp\` to be reachable as an installed package from the directory where the generated stubs will be executed. When stubs live under \`/tmp/jira-mcp/<session>/api/\` (the intended location), no \`node_modules\` exists in any ancestor directory, so Node module resolution would fail for that specifier.

The saving grace: \`Ref\` is only ever imported as \`import type\`, so it is erased at runtime. At tsx/ts-node execution time the broken import path causes no error. It would only bite if someone runs \`tsc --noEmit\` or \`tsc\` against the generated files without \`noResolve\`.

That said, the current state creates a subtle trap for PR #10 (which calls \`generateApi()\` from server startup): if it forgets to pass \`refsImportPath\`, the generated \`types.ts\` will be technically wrong for anyone trying to type-check the stubs. A short JSDoc on \`DEFAULT_REFS_IMPORT\` noting "This path only works if jira-mcp is a resolvable package; PR #10 must pass an absolute path" would make the intention explicit and prevent a future developer from trusting the default by accident.

scottlepp:
No blockers. Three suggestions found:

**generator.ts:57** (suggestion) — `DEFAULT_REFS_IMPORT = "jira-mcp/build/types/refs.js"` is a package-relative specifier that cannot resolve from `/tmp/jira-mcp/<session>/api/` at tsc type-check time. At tsx runtime it is harmless (all `Ref` imports are `import type` and are erased), but PR #10 must pass an explicit absolute path or the generated `types.ts` will be wrong for anyone who tries to type-check the stubs. The comment on the constant should say so explicitly.

**generator.ts:102** (suggestion) — `planApi` pushes one `PlannedFile` entry per operation with no duplicate-path guard. If two operations ever share the same `category/verb` key, `generateApi` silently overwrites the first stub with the second while the `stubPaths.length === operations.length` integration test still passes (it counts plan entries, not unique paths). The current 85-op manifest has no collisions (verified), but a one-line `Set` guard and a matching test would make the invariant enforceable.

**tests/codeapi/generator.test.ts:183** (suggestion) — No test exercises a reserved-word verb. Nine real operations use verb `delete`, which `safeIdentifier` correctly renames to `delete_`. The category index re-export form (`export { delete_ } from "./delete.js"`) is also correct — verified against the built output. But neither `renderStub` nor `renderCategoryIndex` has a unit test asserting this rename, so a future refactor of `safeIdentifier` or the re-export template could silently introduce a JS syntax error in generated code.

Everything else checked out:
- All nine categories of `delete` operations generate syntactically valid stubs exporting `delete_`.
- `issueLink/types.ts` (stub for `issueLink.types`) does not collide with the root `types.ts`; the relative import `../types.js` inside a category stub resolves correctly.
- `watcher.add`'s `rawString` bodyShape is a dispatcher concern and needs no change in the generator — the stub correctly passes `args` through `invoke` and leaves wire-format handling to the bridge.
- `planApi` sort is deterministic for the all-ASCII manifest paths; `localeCompare` ordering of `_` vs lowercase letters is consistent across Node environments.
- The `_client.ts` placeholder is compatible with PR #9's integration: `renderClientFile()` can be extended to accept a socket path without touching the stub or category-index templates.
- Type signatures are appropriately permissive: `string | number` for path/query, `unknown` for body, and typed args interfaces are assignable to `Record<string, unknown>` as `invoke` requires.

PR #177, opened 2026-04-28
v2: code-api IPC bridge (PR #9)

scottlepp:
## Summary
- Real bridge that backs the placeholder `_client.ts` shipped in PR #8. The MCP server runs a local socket listener; generated stubs in the agent's tsx execution context connect, send a one-line JSON request, and get back a `SandboxResult` whose `ref` points at the full response on disk and whose `summary` carries the trimmed projection.
- Transport: Unix domain socket on POSIX (under the session cache dir), loopback TCP on Windows. Falls back to a hashed short path under `tmpdir` when the cache path would exceed macOS's ~104-byte `sun_path` cap.
- Wire format: newline-delimited JSON, `{ id, method: "invoke", params: { operation, args } }` requests; `{ id, result }` or `{ id, error }` responses.

## What changed
- **`src/codeapi/bridge.ts`** (new) — server side. `startBridge({ manifest, client })` returns a `BridgeServer` with `address` (string for `JIRA_MCP_SOCKET`) and `close()`. Per-connection ND-JSON loop dispatches each request through `invokeAndSandbox`, which calls `invokeOperationRaw` and then sandboxes the raw response with the trim projection feeding `summary`.
- **`src/codeapi/templates.ts` (`renderClientFile`)** — replaces the throwing placeholder with a real socket client. Type-checks under `@types/node` strict.
- **`src/core/manifest.ts`** — extracts the wire-call body of `invokeOperation` into `invokeOperationRaw` so the bridge can sandbox before the trim runs. `invokeOperation` is now a thin wrapper that calls `invokeOperationRaw` and applies the trim — Layer 2 behavior unchanged. Adds a small `findOperation` helper.
- **`tests/codeapi/bridge.test.ts`** (new) — 11 tests. Covers `invokeAndSandbox` in isolation, end-to-end happy path over a real socket, unknown-operation error, protocol error (malformed method), `close()` cleanup, and stale-socket recovery.
- **`tests/codeapi/generator.test.ts`** — updated assertions on the new client template.

## Out of scope (next PR)
PR #10 wires `startBridge` into the MCP server's startup path, decides when to call `generateApi`, and sets `JIRA_MCP_SOCKET` on the env when the agent spawns its execution context.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 353 passed, 1 skipped (PR #11 placeholder)
- [x] Emitted `_client.ts` type-checks under `--strict --types node`
- [x] macOS `sun_path` overflow handled — bridge falls back to hash-shortened path; behavior covered by tests

### DIFF
--- a/src/codeapi/bridge.ts
+++ b/src/codeapi/bridge.ts
@@ -148,14 +148,19 @@ export async function startBridge(
 ): Promise<BridgeServer> {
   const addr = opts.address ?? defaultBridgeAddress();
 
-  // POSIX: ensure the socket dir exists, and remove a stale socket
-  // file from a prior crashed session before binding (otherwise
-  // `listen()` errors with EADDRINUSE even though no process is
-  // holding it). Windows TCP doesn't need this.
+  // POSIX: ensure the socket dir exists, and remove a stale entry
+  // from a prior crashed session before binding (otherwise `listen()`
+  // errors with EADDRINUSE even though no process is holding it).
+  // `recursive: true` covers the case where something else has
+  // pre-created a *directory* at the socket path — without it, `rm`
+  // would throw EISDIR and kill startup. The hash-fallback path
+  // under tmpdir is predictable from the session id, so this is a
+  // hardening measure against same-host availability attacks. Windows
+  // TCP doesn't need this branch at all.
   if (addr.listen.path) {
     await fs.mkdir(path.dirname(addr.listen.path), { recursive: true });
     if (existsSync(addr.listen.path)) {
-      await fs.rm(addr.listen.path, { force: true });
+      await fs.rm(addr.listen.path, { force: true, recursive: true });
     }
   }
 
--- a/src/codeapi/bridge.ts
+++ b/src/codeapi/bridge.ts
@@ -1,0 +1,337 @@
+// Layer 3 IPC bridge (PR #9).
+//
+// The MCP server opens a local socket; generated stubs in the agent's
+// execution environment connect to it and forward each call as a
+// newline-delimited JSON request. The server resolves the operation
+// against the manifest, runs it through `invokeOperationRaw`, and
+// sandboxes the response — returning a `SandboxResult` whose `ref`
+// points at the full JSON on disk and whose `summary` carries the
+// trimmed projection.
+//
+// Wire format (one JSON object per line, in both directions):
+//
+//   request:  { "id": "<string>", "method": "invoke",
+//               "params": { "operation": "<name>",
+//                           "args": <object> } }
+//   ok:       { "id": "<string>", "result": <SandboxResult> }
+//   err:      { "id": "<string>",
+//               "error": { "name": "<string>", "message": "<string>" } }
+//
+// The `id` field lets multiple in-flight requests on one connection
+// be demultiplexed by the client. Keeping a small explicit envelope
+// (rather than the full JSON-RPC 2.0 spec) means we don't carry
+// notification/batch baggage we don't use.
+//
+// Transport: Unix domain socket on POSIX, loopback TCP on Windows.
+// `node:net` handles both shapes. Socket addresses surface to the
+// agent's execution context via the `JIRA_MCP_SOCKET` env var (set
+// at spawn time by PR #10).
+
+import * as net from "node:net";
+import * as fs from "node:fs/promises";
+import * as os from "node:os";
+import * as path from "node:path";
+import { createHash } from "node:crypto";
+import { existsSync } from "node:fs";
+
+import type { JiraClient } from "../auth/jira-client.js";
+import {
+  invokeOperationRaw,
+  type Manifest,
+} from "../core/manifest.js";
+import { sandbox, sessionCacheDir } from "../core/sandbox.js";
+import { trimRegistry } from "../core/trim-registry.js";
+import type { SandboxResult } from "../types/refs.js";
+
+// --- Wire shapes ------------------------------------------------------
+
+export interface BridgeRequest {
+  id: string;
+  method: "invoke";
+  params: {
+    operation: string;
+    args?: Record<string, unknown>;
+  };
+}
+
+export interface BridgeOkResponse {
+  id: string;
+  result: SandboxResult<unknown>;
+}
+
+export interface BridgeErrResponse {
+  id: string;
+  error: {
+    name: string;
+    message: string;
+  };
+}
+
+export type BridgeResponse = BridgeOkResponse | BridgeErrResponse;
+
+// --- Address resolution -----------------------------------------------
+
+// Where the server listens. POSIX uses a Unix domain socket inside
+// the session cache dir (auto-cleaned with the rest of the session);
+// Windows lacks Unix sockets, so we fall back to loopback TCP.
+//
+// Returns a `net.ListenOptions`-shaped object plus a stable string
+// representation suitable for setting in `JIRA_MCP_SOCKET`. The
+// string is what the client uses to reconnect; the listen options
+// are what `net.createServer().listen()` consumes.
+export interface BridgeAddress {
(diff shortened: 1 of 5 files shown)


src/codeapi/templates.ts
```diff
@@ -223,24 +223,102 @@ export function renderTypesFile(refsImportPath: string): string {
...
+    `let nextId = 0;\n\n` +
+    `export function invoke(\n` +
+    `  operation: string,\n` +
+    `  args: Record<string, unknown>,\n` +
     `): Promise<Ref<unknown>> {\n` +
-    `  throw new Error(\n` +
-    `    \`jira-mcp code-api bridge not implemented yet (operation: \${operationName}). \` +\n` +
-    `      \`Wired in PR #9 — until then, use the classic MCP tools.\`,\n` +
-    `  );\n` +
+    `  const target = resolveSocket();\n` +
```
scottlepp: **Bug: `invoke()` throws synchronously instead of returning a rejected Promise**

`resolveSocket()` is called outside the `new Promise(...)` constructor (line 275 of the rendered template source). When `JIRA_MCP_SOCKET` is missing or malformed, `resolveSocket()` throws synchronously — so `invoke()` itself throws rather than returning a rejected Promise.

Any caller that follows the promise chain without `await` will get an unhandled exception:

```ts
// This silently crashes — .catch() is never reached because the sync throw
// escapes before a Promise is returned.
getIssue("PROJ-1").catch(err => handleError(err));
```

`await getIssue(...)` inside an `async` function still works (the `async` wrapper catches sync throws), but it is fragile: callers have no way to know this invariant holds, and `Promise.all([...]).catch(...)` patterns will also explode.

Fix: move `resolveSocket()` inside the `new Promise<Ref<unknown>>((resolve, reject) => {` constructor so any throw is caught and turned into a rejection:

```ts
return new Promise<Ref<unknown>>((resolve, reject) => {
  let target;
  try {
    target = resolveSocket();
  } catch (err) {
    reject(err);
    return;
  }
  const socket: Socket = connect(target as never);
  // ...
});
```

tests/codeapi/bridge.test.ts
```diff
@@ -0,0 +1,349 @@
...
+      message: expect.stringContaining("missing.op"),
+    });
+  });
+
+  it("returns a ProtocolError when the request shape is malformed", async () => {
+    const { client } = makeMockClient();
+    bridge = await startBridge({ manifest: fixtureManifest, client });
+
+    const resp = await new Promise<any>((resolve, reject) => {
+      const socket = net.connect({ path: bridge!.address });
```
scottlepp: **Test fails on Windows: `net.connect({ path: bridge!.address })` ignores TCP address format**

This test constructs the socket connection directly with `{ path: bridge!.address }` instead of going through the `callBridge()` helper. On Windows, `bridge!.address` is `"tcp:127.0.0.1:<port>"` — a TCP address string, not a filesystem path. Passing it as `{ path: "tcp:127.0.0.1:..." }` to `net.connect` will fail.

All the other multi-transport tests use `callBridge()`, which correctly parses both address formats. Either swap this for `callBridge()`, or add the same `if (process.platform === "win32") return;` guard used by the other POSIX-only tests in this suite.

src/codeapi/bridge.ts
```diff
@@ -0,0 +1,337 @@
...
+  const addr = opts.address ?? defaultBridgeAddress();
+
+  // POSIX: ensure the socket dir exists, and remove a stale socket
+  // file from a prior crashed session before binding (otherwise
+  // `listen()` errors with EADDRINUSE even though no process is
+  // holding it). Windows TCP doesn't need this.
+  if (addr.listen.path) {
+    await fs.mkdir(path.dirname(addr.listen.path), { recursive: true });
+    if (existsSync(addr.listen.path)) {
+      await fs.rm(addr.listen.path, { force: true });
```
scottlepp: **DoS: `fs.rm({ force: true })` without `{ recursive: true }` fails on a pre-created directory**

You already flagged the `/tmp` pre-creation concern in your questions; the risk is slightly different from what you described. `force: true` alone suppresses ENOENT but does **not** suppress `EISDIR`. If an attacker pre-creates a **directory** at the socket path (rather than a file), `fs.rm({ force: true })` throws:

```
[Error: Path is a directory: rm returned EISDIR (is a directory) /tmp/jmcp-<hash>.sock]
```

This propagates as a rejected promise from `startBridge()`, killing server startup.

The hash fallback path under `os.tmpdir()` is the more exposed case — an attacker who knows the session ID (e.g., because `MCP_SESSION_ID` is reused or predictable) can precompute the hash and pre-create the directory before the server starts. The preferred path under `sessionCacheDir()` requires the attacker to also pre-create the session directory, which is harder but possible in the same window.

Fix: add `{ recursive: true }` to the stale-socket cleanup:

```ts
await fs.rm(addr.listen.path, { force: true, recursive: true });
```

`recursive: true` was added in Node 16.x and unconditionally removes the target whether it is a file, symlink, or directory.

scottlepp:
**3 line comments: 1 blocker, 1 bug, 1 security suggestion.**

**Blocker — `src/codeapi/templates.ts` line 275**
`resolveSocket()` is called synchronously before the `new Promise(...)` constructor in the generated `invoke()`. When `JIRA_MCP_SOCKET` is missing or malformed, the function throws instead of returning a rejected Promise. `await invoke(...)` inside an `async` function still works (the async wrapper catches sync throws), but any call site using `.catch()` directly — including `Promise.all([...]).catch(...)` — gets an unhandled exception. Fix: move the `resolveSocket()` call inside the Promise constructor with a try/catch that calls `reject()`.

**Bug — `tests/codeapi/bridge.test.ts` line 290**
The ProtocolError test uses `net.connect({ path: bridge!.address })` unconditionally. On Windows, `bridge!.address` is `"tcp:127.0.0.1:<port>"`, not a socket path. The test would fail. All other multi-transport tests go through `callBridge()`, which handles both formats. Fix: use `callBridge()` here, or add a `if (process.platform === "win32") return;` guard.

**Security suggestion — `src/codeapi/bridge.ts` line 158**
`fs.rm(path, { force: true })` suppresses ENOENT but not EISDIR. If an attacker pre-creates a *directory* at the socket path, `startBridge()` throws with EISDIR and the server fails to start. The hash-fallback path under `os.tmpdir()` is the exposed vector (predictable from the session ID). Fix: add `{ recursive: true }` to the `fs.rm` call.

---

**On the questions you raised:**

- **Concurrent in-flight requests / OOO responses:** the design is correct. The server tags each response with the request's `id`, so a future multiplexing client can demux. The current per-call connection model is safe today. No issue.

- **UNIX_SOCKET_PATH_MAX = 100:** conservative but sound. The hash-fallback path on macOS measures ~71 bytes on a typical system (tmpdir is ~48 chars + `/jmcp-` + 12 hex chars + `.sock`), well under the limit. No issue — though see the EISDIR comment above for the related DoS.

- **Error shape (JiraError fields lost at wire boundary):** `statusCode` and response body are indeed stripped. That's a reasonable trade-off for now given the agent can re-fetch via the ref; worth a TODO comment but not a blocker.

- **`invokeOperationRaw` refactor correctness:** safe. `invokeOperation` is now a thin wrapper that calls `invokeOperationRaw` then applies trim — same sequence as before. The existing manifest test "applies trim projection when specified" exercises the full path. No regression risk.

- **tsx environment / module resolution:** the generated `_client.ts` only imports `node:net` and a relative `./types.js`. Both resolve identically in tsx and tsc. No issue.

scottlepp:
3 line comments posted: 1 blocker (sync throw from invoke()), 1 bug (Windows test failure), 1 security suggestion (EISDIR DoS). Full details in line comments. See also inline responses to your design questions: concurrent in-flight requests, UNIX_SOCKET_PATH_MAX, invokeOperationRaw refactor, and tsx compatibility are all fine.

PR #178, opened 2026-04-28
v2: wire code-api as default tool mode (PR #10)

scottlepp:
## Summary
Layer 3 ("code-api" mode) becomes the default surface. At boot the MCP server now generates the TypeScript stubs and starts the IPC bridge, then exposes a single tool — `jira_code_api` — that hands the agent the on-disk API path. `JIRA_TOOL_MODE=classic` falls back to the 16 consolidated v2 tools.

This is the PR that flips the runtime token cost: tool-list size goes from ~3-4k tokens (16 tools) to under 500 tokens (one tool with a tight description), and per-call response shape goes from raw Jira JSON to a Ref<T> with a trimmed `summary` plus a filesystem `ref` for the full body.

## What changed
- **`src/config.ts`** — adds `toolMode` (`"code-api"` | `"classic"`), parsed from `JIRA_TOOL_MODE`. Empty/unset defaults to `"code-api"`. Unknown values throw with the offending value in the message.
- **`src/codeapi/tool.ts`** (new) — definition + response builder for the `jira_code_api` tool. Description under 500 chars; response body under 1KB.
- **`src/codeapi/boot.ts`** (new) — extracted `bootCodeApi()`: runs the cleanup-stale-sessions sweep, generates stubs to `${sessionCacheDir}/api`, starts the bridge, publishes `JIRA_MCP_SOCKET` to the server's env so child processes (Bash, tsx) inherit it. Has test seams via `apiDir` / `refsImportPath` / `cleanupSessions` overrides.
- **`src/index.ts`** — branches on `cfg.toolMode` at startup, populates `modeState`, and the request handlers read from it. SIGINT/SIGTERM closes the bridge in code-api mode. Existing classic dispatch + resources surface unchanged.
- **Tests** — `tests/config.test.ts` (toolMode parsing), `tests/codeapi/tool.test.ts` (definition + response shape), `tests/codeapi/boot.test.ts` (full stack: boot → drive bridge over real socket → assert SandboxResult round-trip).

## Out of scope
PR #11 covers docs (README rewrite, migration guide, token-budget benchmark).

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 367 passed, 1 skipped (PR #11 placeholder); +12 new
- [x] Smoke test default mode: server boots, generates 85+ stubs under `${sessionCacheDir}/api/`, opens `ipc.sock`, logs the api path
- [x] Smoke test classic mode (`JIRA_TOOL_MODE=classic`): server boots, no api dir / socket created, only `mode=classic` logged
- [x] SIGTERM cleanup: bridge closes gracefully, socket file removed

### DIFF
--- a/src/codeapi/boot.ts
+++ b/src/codeapi/boot.ts
@@ -1,0 +1,76 @@
+// Code-api startup glue (PR #10).
+//
+// Pulled out of `src/index.ts` so the wiring (cleanup → generate
+// stubs → start bridge → publish socket env var) is testable in
+// isolation, without standing up the MCP transport.
+
+import * as path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import type { JiraClient } from "../auth/jira-client.js";
+import { operations } from "../core/operations.js";
+import { cleanupStaleSessions, sessionCacheDir } from "../core/sandbox.js";
+import { startBridge, type BridgeServer } from "./bridge.js";
+import { generateApi } from "./generator.js";
+import type { CodeApiToolContext } from "./tool.js";
+
+// Where the generator points stubs' `types.ts` to find Ref<T>.
+// Stubs live in the session cache dir, far from the installed
+// jira-mcp package, so we use the absolute path of this build's
+// compiled refs.js. Resolved relative to *this* module so it stays
+// correct under `npm link`, npm install, or running from `build/`.
+export function defaultRefsImportPath(): string {
+  const here = path.dirname(fileURLToPath(import.meta.url));
+  // src/codeapi/boot.ts → src/types/refs.ts (after compile:
+  // build/codeapi/boot.js → build/types/refs.js).
+  return path.join(here, "..", "types", "refs.js");
+}
+
+export interface BootedCodeApi {
+  bridge: BridgeServer;
+  ctx: CodeApiToolContext;
+}
+
+export interface BootCodeApiOpts {
+  client: JiraClient;
+  // Override hooks for tests. Production callers leave these unset.
+  refsImportPath?: string;
+  apiDir?: string;
+  // When false (default true) skip the cleanup-stale-sessions sweep.
+  // Tests turn this off to avoid touching siblings of their session
+  // dir.
+  cleanupSessions?: boolean;
+}
+
+export async function bootCodeApi(
+  opts: BootCodeApiOpts,
+): Promise<BootedCodeApi> {
+  if (opts.cleanupSessions !== false) {
+    await cleanupStaleSessions().catch(() => {
+      // Best-effort. Permission failures on a stale dir shouldn't
+      // block startup.
+    });
+  }
+
+  const apiDir = opts.apiDir ?? path.join(sessionCacheDir(), "api");
+  await generateApi({
+    manifest: operations,
+    outDir: apiDir,
+    refsImportPath: opts.refsImportPath ?? defaultRefsImportPath(),
+  });
+
+  const bridge = await startBridge({
+    manifest: operations,
+    client: opts.client,
+  });
+
+  // Place the socket in the *server* env so any subprocess (Claude
+  // Code's Bash tool, a tsx invocation) inherits it without the
+  // user having to configure anything.
+  process.env.JIRA_MCP_SOCKET = bridge.address;
+
+  return {
+    bridge,
+    ctx: { apiDir, socketAddress: bridge.address },
+  };
+}
--- a/src/codeapi/boot.ts
+++ b/src/codeapi/boot.ts
@@ -19,10 +19,16 @@ import type { CodeApiToolContext } from "./tool.js";
 // jira-mcp package, so we use the absolute path of this build's
 // compiled refs.js. Resolved relative to *this* module so it stays
 // correct under `npm link`, npm install, or running from `build/`.
+//
+// Caveat: this targets compiled output (build/codeapi/boot.js →
+// build/types/refs.js). If anyone ever runs the server through `tsx
+// src/index.ts` directly, the path resolves to `src/types/refs.js`
+// — which doesn't exist (only refs.ts does). The repo has no tsx
+// dev script today, so production paths are safe; the npm `dev`
+// script is `tsc --watch`, which writes refs.js to build/. Callers
+// running outside that flow should pass an explicit refsImportPath.
 export function defaultRefsImportPath(): string {
   const here = path.dirname(fileURLToPath(import.meta.url));
-  // src/codeapi/boot.ts → src/types/refs.ts (after compile:
-  // build/codeapi/boot.js → build/types/refs.js).
   return path.join(here, "..", "types", "refs.js");
 }
 
--- a/src/codeapi/tool.ts
+++ b/src/codeapi/tool.ts
@@ -51,12 +51,20 @@ export interface CodeApiToolResponse {
 export function buildCodeApiToolResponse(
   ctx: CodeApiToolContext,
 ): CodeApiToolResponse {
+  // The agent typically runs this via Claude Code's Bash tool, whose
+  // child shells *do not* inherit env vars from the MCP server
+  // process. So the snippet must export JIRA_MCP_SOCKET inline rather
+  // than assume it's already set. tsx is the recommended runner — it
+  // executes the .ts stubs directly.
   const usage = [
-    `// In your shell (tsx required):`,
+    `# Run in your shell (tsx required). The JIRA_MCP_SOCKET prefix is`,
+    `# load-bearing — child shells don't inherit the MCP server's env.`,
+    `JIRA_MCP_SOCKET=${ctx.socketAddress} npx tsx -e '`,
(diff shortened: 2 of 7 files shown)


src/index.ts
```diff
@@ -17,136 +18,162 @@ import {
...
+      ")",
+  );
 }
 
+async function shutdown(signal: NodeJS.Signals) {
+  console.error(`Received ${signal}, shutting down`);
+  if (modeState?.mode === "code-api") {
+    await modeState.bridge.close().catch(() => {});
+  }
+  process.exit(0);
```
scottlepp, line 182: **Suggestion — shutdown skips MCP transport drain and undici pool**

`shutdown()` calls `bridge.close()` then immediately `process.exit(0)`, but leaves two resources open:

1. The MCP `Server` / `StdioServerTransport`. `server.close()` exists on the `Protocol` base class and calls `transport.close()`, which flushes the write buffer and emits the `onclose` event. Skipping it means any in-progress MCP response being assembled in the event loop at the moment SIGTERM arrives can be silently truncated.

2. The undici connection pool. `closeHttpPool()` is exported from `src/core/http.ts` specifically for this purpose (its comment says "Exposed for graceful shutdown if the host process ever needs it"). Omitting it means up to 8 keep-alive TLS connections are closed by the OS rather than drained — not a correctness problem in practice, but it does mean a pending Jira HTTP call (e.g. a slow POST) can be cut mid-stream.

Suggested fix for `shutdown()`:

```ts
async function shutdown(signal: NodeJS.Signals) {
  console.error(`Received ${signal}, shutting down`);
  if (modeState?.mode === "code-api") {
    await modeState.bridge.close().catch(() => {});
  }
  await server.close().catch(() => {});   // flush MCP transport
  await closeHttpPool().catch(() => {});  // drain undici keep-alive pool
  process.exit(0);
}
```

Neither omission causes data loss under normal operation (the OS cleans up file descriptors), but the exported `closeHttpPool` hook exists precisely for this call site.

src/codeapi/boot.ts
```diff
@@ -0,0 +1,76 @@
...
+// Stubs live in the session cache dir, far from the installed
+// jira-mcp package, so we use the absolute path of this build's
+// compiled refs.js. Resolved relative to *this* module so it stays
+// correct under `npm link`, npm install, or running from `build/`.
+export function defaultRefsImportPath(): string {
+  const here = path.dirname(fileURLToPath(import.meta.url));
+  // src/codeapi/boot.ts → src/types/refs.ts (after compile:
+  // build/codeapi/boot.js → build/types/refs.js).
+  return path.join(here, "..", "types", "refs.js");
+}
```
scottlepp, line 33: **Suggestion — `defaultRefsImportPath()` produces a non-existent path when the server is run via `tsx src/index.ts`**

The comment says the path is correct under `npm link`, `npm install`, and `running from build/`. That's accurate. But if a developer runs the server directly from source with `tsx src/index.ts` (a natural workflow when iterating on the generator), `import.meta.url` resolves to `file:///…/src/codeapi/boot.ts`, which makes `here` = `src/codeapi/` and the returned path = `src/types/refs.js`. That file doesn't exist — only `src/types/refs.ts` does. Every generated stub that imports `Ref<T>` from that path will fail to resolve.

This only bites the tsx-from-source developer path, not the npm install / npm link path. If tsx-from-source isn't a supported workflow (there's no `tsx` dev script in `package.json`) you can close this by updating the comment to say so explicitly. If you do want to support it, the fix is to detect whether we're running from a `.ts` file and swap the extension:

```ts
export function defaultRefsImportPath(): string {
  const here = path.dirname(fileURLToPath(import.meta.url));
  const candidate = path.join(here, "..", "types", "refs.js");
  // When running via tsx the module lives at *.ts; refs.ts exists but refs.js doesn't.
  // Fall back to refs.ts so the import resolves in both compiled and source runs.
  const tsCandidate = candidate.replace(/\.js$/, ".ts");
  return existsSync(candidate) ? candidate : tsCandidate;
}
```

(Uses the sync `existsSync` that's already imported in `bridge.ts` — boot.ts would need to import it too.)

src/codeapi/tool.ts
```diff
@@ -0,0 +1,69 @@
...
+  usage: string;
+}
+
+export function buildCodeApiToolResponse(
+  ctx: CodeApiToolContext,
+): CodeApiToolResponse {
+  const usage = [
+    `// In your shell (tsx required):`,
+    `import * as jira from "${ctx.apiDir}/index.js";`,
+    `const issue = await jira.issue.get({ issueIdOrKey: "PROJ-1" });`,
```
scottlepp, line 64: **Suggestion — usage snippet doesn't show how to set `JIRA_MCP_SOCKET` before running tsx**

The response includes `socketEnv: "JIRA_MCP_SOCKET"` and `socketAddress: "…/ipc.sock"` as separate fields, but the `usage` snippet shows only:

```ts
// In your shell (tsx required):
import * as jira from "/path/to/api/index.js";
const issue = await jira.issue.get({ issueIdOrKey: "PROJ-1" });
```

An agent that follows this snippet verbatim will run `tsx script.ts` without setting `JIRA_MCP_SOCKET`, hitting the error: `"jira-mcp bridge address not set: JIRA_MCP_SOCKET env var missing"`. The agent has to infer from the separate `socketEnv`/`socketAddress` fields that it must prefix the invocation — that's not stated anywhere.

Simplest fix: embed the env-var setup directly in the snippet so it's copy-pasteable:

```ts
const usage = [
  `// In your shell (tsx required):`,
  `// export JIRA_MCP_SOCKET=${ctx.socketAddress}`,
  `// or inline: JIRA_MCP_SOCKET=${ctx.socketAddress} tsx script.ts`,
  `import * as jira from "${ctx.apiDir}/index.js";`,
  `const issue = await jira.issue.get({ issueIdOrKey: "PROJ-1" });`,
  ...
].join("\n");
```

Alternatively, one sentence in the tool description — `"set JIRA_MCP_SOCKET to the returned socketAddress before invoking tsx"` — would also close the gap.

scottlepp:
No blockers. 3 suggestions posted as line comments; answers to the specific questions below.

**Line comments posted**
- `src/index.ts:171` — `shutdown()` skips `server.close()` and `closeHttpPool()`. The undici pool hook was exported specifically for this; omitting it means in-flight Jira HTTP calls can be killed mid-stream on SIGTERM.
- `src/codeapi/boot.ts:27` — `defaultRefsImportPath()` emits `src/types/refs.js` when the server is started via `tsx src/index.ts`. That file doesn't exist under `src/` (only `refs.ts` does). Not a bug for the npm install / npm link path; close it by clarifying the comment or adding a `.ts` fallback.
- `src/codeapi/tool.ts:57` — The `usage` snippet doesn't show `JIRA_MCP_SOCKET=<addr>` before the `tsx` invocation. The agent has `socketEnv`/`socketAddress` as separate JSON fields, but nothing connecting them to the shell invocation. Stubs throw a clear error if the var is missing, so this surfaces quickly in practice — still worth fixing before agents start writing scripts from the snippet.

**Answers to the specific questions**

_Lifecycle correctness (SIGINT/SIGTERM):_ The bridge is correctly closed before `process.exit(0)`, which is the most important step (socket file gets removed). The MCP transport (`server.close()`) and undici pool (`closeHttpPool()`) are left open. The transport omission is low-risk because stdio is not a real socket and the OS drains it on exit; the undici omission matters more if there's a pending POST at the moment the signal lands. Both are straightforward one-liners to add.

_`defaultRefsImportPath()` correctness:_ Correct for every install shape where the server runs as compiled JS (`build/codeapi/boot.js` → `build/types/refs.js` — verified locally). Breaks only under `tsx src/index.ts`. The comment accurately lists the supported shapes; the tsx-from-source case isn't listed and isn't supported by any package.json script, so this is a documentation gap rather than a production bug.

_Stale `JIRA_MCP_SOCKET` in a child shell:_ Not a real problem. `process.env` assignment in the MCP server process overwrites the value for that process and any children it forks — the new value is always what `bootCodeApi` sets. More relevantly, Claude Code's Bash tool runs as a child of Claude Code, not of the MCP server, so `JIRA_MCP_SOCKET` doesn't propagate via fork at all. The correct delivery channel is the tool response fields (`socketAddress`), which is what the comment on line 64–70 says. No issue here.

_Tool description discoverability:_ The description correctly signals "call once" and "then import the stubs". The gap is not in the description but in the `usage` snippet, which omits the env-var prefix. See the line comment on `tool.ts:57`.

_`modeState` race / "throw if not initialized":_ The guard is sufficient. Node's event loop is single-threaded; `server.connect(transport)` starts processing stdin only after it resolves, and `modeState` is written before `server.connect()` is called in `main()`. An MCP request cannot arrive before `main()` completes the assignment.

_Duplicate socket client (`boot.test.ts` test 2 vs `bridge.test.ts`):_ Crosses the duplication threshold by roughly 40 lines of identical address-parsing + read-loop logic. Worth extracting to a shared `tests/helpers/bridge-client.ts`, but it's test-only and doesn't affect correctness. Agree this can be deferred to a cleanup PR.

PR #179, opened 2026-04-28
v2: token-budget benchmark script (PR #11 prep)

scottlepp:
## Summary
A manual-run benchmark that spawns the real MCP server in each tool mode, talks MCP to it via the stdio client, and measures how many bytes of request/response would land in an agent's context. Hits real Jira so the numbers reflect what an agent actually sees in production. Output is a markdown table — the table that PR #11 will paste into the v2.0 CHANGELOG.

## Numbers from a real run

Against my Jira instance (rich ticket = 100+ comments, several attachments; simple = no comments):

**Tool-list cost (one-time, every session):**
- classic: 30.6KB / ~7800 tokens
- code-api: 378B / ~95 tokens
- **83× smaller in code-api**

**Per-flow cost (`code-api summary` = agent only reads inline summary; `code-api +full` = upper bound, agent reads every ref):**

| scenario | classic | code-api summary | code-api +full | summary vs classic |
| --- | --- | --- | --- | --- |
| fetch 1 simple ticket | 1.4KB | 1.3KB | 43.4KB | 1.1× smaller |
| investigate rich ticket | 141.9KB | 85.1KB | 499.9KB | 1.7× smaller |
| JQL search ~10 tickets | 5.0KB | 3.6KB | 28.7KB | 1.4× smaller |

So code-api's headline win is the **tool-list cost**, not per-call. The summary-only path beats classic by ~1.1–1.7×; the +full path is *worse* than classic if the agent reads every ref. Numbers feed into the "should code-api stay default" discussion for PR #11.

The rich-ticket classic number (142KB) is also unexpectedly high — issue + comment trims should compress harder. Suggests a separate trim bug (filed as a follow-up).

## What's new
- `scripts/benchmark.ts` — script with three scenarios. Uses `@modelcontextprotocol/sdk`'s Client + StdioClientTransport so the wire bytes are what an agent actually sees. For code-api, drives the bridge directly (same wire format the generated `_client.ts` uses).
- `package.json` — adds `tsx` as a devDep + `npm run benchmark`.
- `.env.local` carries `JIRA_BENCH_TICKET_RICH` / `JIRA_BENCH_TICKET_SIMPLE` plus normal Jira creds; not committed.

## Out of scope
- v1 baseline numbers (deferred — the v2-vs-v2 comparison is what drives the "default mode" decision).
- CI integration. Manual-only by design; benchmarks against live Jira don't belong in CI.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 368 passed (no source changes outside scripts/)
- [x] `npm run benchmark` runs against real Jira, produces the table above

### DIFF
--- a/scripts/benchmark.ts
+++ b/scripts/benchmark.ts
@@ -1,0 +1,520 @@
+// Token-budget benchmark for v2 (PR #11 prep).
+//
+// Spawns the real MCP server in each tool mode, talks MCP to it via the
+// stdio client, and measures how many bytes of request/response would
+// land in an agent's context for a fixed sequence of Jira calls. Hits
+// real Jira — no fixtures — so the numbers reflect what an agent
+// actually sees in production.
+//
+// Usage:
+//   JIRA_BENCH_TICKET_RICH=ABC-1 JIRA_BENCH_TICKET_SIMPLE=ABC-2 \
+//     npm run benchmark
+//
+// Requires .env.local (or shell env) with JIRA_HOST / JIRA_EMAIL /
+// JIRA_API_TOKEN. The two ticket env vars name a "rich" ticket
+// (comments + attachments — where v2 should win) and a "simple"
+// ticket (no comments — where v2 likely doesn't help).
+//
+// Output: a markdown table on stdout. Each row is one scenario; each
+// numeric cell is bytes that would enter the agent's context window.
+// For code-api mode we report two numbers (summary-only and
+// summary+full) since the cost depends on whether the agent decides
+// to read each ref's full body.
+
+import { promises as fs, readFileSync } from "node:fs";
+import * as net from "node:net";
+import * as path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js";
+
+// --- Config -----------------------------------------------------------
+
+interface BenchEnv {
+  JIRA_HOST: string;
+  JIRA_EMAIL: string;
+  JIRA_API_TOKEN: string;
+  JIRA_BENCH_TICKET_RICH: string;
+  JIRA_BENCH_TICKET_SIMPLE: string;
+}
+
+function loadEnvLocal(): void {
+  // The MCP server itself loads .env.local via dotenv at startup.
+  // The benchmark script also needs the env locally so it can read
+  // ticket keys for scenario assembly. Keep it simple: read the
+  // file directly, set vars that aren't already set.
+  const here = path.dirname(fileURLToPath(import.meta.url));
+  const file = path.resolve(here, "..", ".env.local");
+  let text: string;
+  try {
+    text = readFileSync(file, "utf8");
+  } catch {
+    return; // not present — caller must have set env elsewhere
+  }
+  for (const line of text.split("\n")) {
+    const trimmed = line.trim();
+    if (!trimmed || trimmed.startsWith("#")) continue;
+    const eq = trimmed.indexOf("=");
+    if (eq < 0) continue;
+    const key = trimmed.slice(0, eq);
+    const val = trimmed.slice(eq + 1);
+    if (process.env[key] === undefined) process.env[key] = val;
+  }
+}
+
+function requireEnv(): BenchEnv {
+  const need = [
+    "JIRA_HOST",
+    "JIRA_EMAIL",
+    "JIRA_API_TOKEN",
+    "JIRA_BENCH_TICKET_RICH",
+    "JIRA_BENCH_TICKET_SIMPLE",
+  ] as const;
+  const missing = need.filter((k) => !process.env[k]);
+  if (missing.length > 0) {
+    throw new Error(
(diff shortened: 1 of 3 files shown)


scripts/benchmark.ts
```diff
@@ -0,0 +1,520 @@
...
+  summaryPlusFullBytes: number;
+}
+
+async function classicCallCost(
+  client: Client,
+  toolName: string,
+  args: Record<string, unknown>,
+): Promise<CallCost> {
+  const resp = await client.callTool({ name: toolName, arguments: args });
+  const bytes = byteLength(resp);
```
scottlepp: **Measurement asymmetry: classic includes MCP envelope bytes, code-api does not.**

`byteLength(resp)` measures the full MCP `CallToolResult` object — `{"content":[{"type":"text","text":"..."}]}` — which adds roughly 40-50 bytes of wire envelope per call that the code-api path doesn't pay (since `codeApiCallCost` calls `byteLength(result)` on the unwrapped `SandboxResult`).

For the rich-ticket scenario (142 KB vs 85 KB) this is noise. For the simple-ticket scenario (1.4 KB classic vs 1.3 KB code-api) the envelope overhead is non-trivial relative to the 100-byte gap — the "1.1× smaller" headline for that row could be largely envelope, not data compression.

Fix: measure the text payload only, not the envelope, so both modes are measured at the data level:
```ts
const payload = parseToolText(resp);
const bytes = byteLength(payload); // strip the content[] wrapper
```

scripts/benchmark.ts
```diff
@@ -0,0 +1,520 @@
...
+  args: Record<string, unknown>,
+): Promise<{ summary: any; ref: string; fullSize: number; hash: string; fetchedAt: string }> {
+  return new Promise((resolve, reject) => {
+    const target = address.startsWith("tcp:")
+      ? (() => {
+          const lc = address.lastIndexOf(":");
+          return { host: address.slice(4, lc), port: Number(address.slice(lc + 1)) };
+        })()
+      : { path: address };
+    const socket = net.connect(target as never);
```
scottlepp, line 246: **No socket timeout — benchmark can hang indefinitely on a slow or unresponsive Jira call.**

`net.connect()` has no `setTimeout` configured. If the IPC bridge's Jira HTTP request stalls (rate-limit, network blip, server-side timeout), the `data` event never fires, the Promise never settles, and the script hangs with no output and no error — the only recovery is Ctrl-C.

For an interactive measurement tool that hits a live API, this makes failures hard to diagnose. A 30-second timeout is enough to surface the hang with a useful error:
```ts
const socket = net.connect(target as never);
socket.setTimeout(30_000, () => {
  socket.destroy(new Error(`bridge call timed out after 30s (operation=)`));
});
```
The `error` listener below already calls `reject(err)`, so `socket.destroy(new Error(...))` routes through the existing handler.

scottlepp:
No blockers. 2 suggestions (line 186, line 232).

**Line 186 — measurement asymmetry between modes (suggestion):** `classicCallCost` calls `byteLength(resp)` on the full MCP `CallToolResult` object, which includes the `{"content":[{"type":"text","text":"..."}]}` envelope — roughly 40-50 extra bytes per call. `codeApiCallCost` calls `byteLength(result)` on the unwrapped `SandboxResult`, paying no such overhead. For rich-ticket (142 KB vs 85 KB) this is irrelevant noise. For simple-ticket (1.4 KB vs 1.3 KB) the 40-50 byte envelope is a non-trivial share of the 100-byte gap and could be what the "1.1× smaller" headline is actually measuring. Both modes should be compared at the same level (either both include the transport envelope, or neither does).

**Line 232 — no socket timeout (suggestion):** `invokeOverBridge` has no `socket.setTimeout()`. If the IPC bridge's Jira HTTP request stalls, the Promise never settles and the script hangs silently until Ctrl-C. A 30s timeout with `socket.destroy(new Error(...))` routes cleanly through the existing `error` listener.

Everything else checked out: the indexOf-based env parser correctly handles `=` in values; the TCP address parser (`lastIndexOf` + slice) is correct; the two server children are isolated by distinct session IDs; the `disconnect()` try/finally cleanup is sound; the `tsx` peer-dep flag in the lockfile doesn't affect installation since it's also a direct devDep. The measurement philosophy questions in the PR description (bootstrap cost attribution, bytes/4 token estimate, Cost struct layout) are all documented and reasonable for the stated purpose.

scripts/benchmark.ts
scottlepp: Addressed in 9a87c79 — `classicCallCost` now strips the MCP envelope via `parseToolText(resp)` before measuring, so both modes are compared at the data level.

Re-running the benchmark with the strip flipped the per-call story: code-api summary went from 1.0–1.3× smaller than classic to ~1.0–1.2× *larger*, confirming the original asymmetry was hiding the real comparison. Numbers in the PR description are updated.

scripts/benchmark.ts
scottlepp: Addressed in 9a87c79 — `invokeOverBridge` now sets `socket.setTimeout(BRIDGE_CALL_TIMEOUT_MS, ...)` (30s) and destroys the socket with a clear error message when it fires. The destroy routes through the existing `error` listener as suggested.

PR #180, opened 2026-04-28
v2: trim list endpoints (comment.list, board.issues, ...)

scottlepp:
## Summary
Fixes a systemic gap surfaced by the benchmark in PR #179: list endpoints (comment.list, board.issues, etc.) were shipping their full untrimmed Jira responses through the bridge, putting 80KB+ of comment ADF or issue rows directly into agent context. Now they go through a trim projection that emits count + pagination metadata only; the full body still lands on disk via `sandbox()`, so the agent reads the ref + runs their own filter (jq, code, whatever) when they want detail.

## Real-world impact

Same MON-317 ticket the benchmark uses, "investigate rich ticket" scenario:

| | before | after |
| --- | --- | --- |
| classic | 84.8KB | **2.8KB** |
| code-api summary | 85.1KB | **3.1KB** |

~30× smaller for the case the plan called out as v2's win case. Other rows (simple ticket, JQL search) unchanged because they were already well-trimmed. Tool-list cost unchanged.

## Design rationale

The shape: `{ total, startAt, maxResults, truncated }`. No inline items. This is closer in spirit to a CLI — the server points at the data, the agent shapes its own view. Inline items would either be wrong for the agent's question (no fixed shape suits "what did Bob say on July 3rd" *and* "give me a count") or balloon context when they don't actually need them.

`searchSummary` keeps inline issue rows as a justified exception: keys+summaries compress to ~80B/row, and the agent often picks a key out of a result list and follows up.

## Trims added (`src/core/trim.ts`)
- **`paginatedListSummary`** — standard Jira PageBean. Recognizes array keys `comments`/`worklogs`/`issues`/`values`/`groups`. Used by 13 ops including comment.list, worklog.list, board.issues, sprint.issues, epic.issues, board.backlog, board.epics, board.list, sprint.listForBoard, project.list, filter.list, user.bulkGet, group.search, group.members.
- **`bareListSummary`** — bare-array endpoints. Used by project.listComponents, project.listVersions, filter.listFavourite, user.search, user.assignable.
- **`watcherListSummary`** / **`voteListSummary`** — special shapes for `/issue/{key}/watchers` and `/issue/{key}/votes` that carry per-caller flags (isWatching, hasVoted) alongside counts.

## Left untrimmed
Lookup endpoints that return small fixed sets: field.list, field.issueTypes, field.statuses, field.priorities, field.resolutions. Agent typically wants the whole list.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 379 passed (was 368; +11 new for the trim functions)
- [x] Re-ran benchmark from PR #179 — investigate-rich-ticket dropped 84.8KB → 2.8KB / 85.1KB → 3.1KB

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -997,9 +997,9 @@ export const operations: Manifest = [
       { name: "caseInsensitive", role: "query" },
       { name: "maxResults", role: "query" },
     ],
-    // /groups/picker returns { groups: [...], total }; treat as a
-    // paginated list — paginatedListSummary's pageItemCount falls
-    // through to the array length when `groups` isn't a known key.
+    // /groups/picker returns { groups: [...], total }. The
+    // `groups` array key is included in paginatedListSummary's
+    // PageBean recognizer alongside comments/worklogs/issues/values.
     trim: "list",
   },
   {
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -167,6 +167,7 @@ export const operations: Manifest = [
       { name: "orderBy", role: "query" },
       { name: "expand", role: "query" },
     ],
+    trim: "list",
   },
   {
     name: "comment.add",
@@ -249,6 +250,7 @@ export const operations: Manifest = [
       { name: "maxResults", role: "query" },
       { name: "orderBy", role: "query" },
     ],
+    trim: "list",
   },
   {
     name: "project.get",
@@ -305,6 +307,7 @@ export const operations: Manifest = [
     verb: "GET",
     pathTemplate: "/project/{projectIdOrKey}/components",
     params: [{ name: "projectIdOrKey", role: "path", required: true }],
+    trim: "bareList",
   },
   {
     name: "project.createComponent",
@@ -328,6 +331,7 @@ export const operations: Manifest = [
       { name: "projectIdOrKey", role: "path", required: true },
       { name: "expand", role: "query" },
     ],
+    trim: "bareList",
   },
   {
     name: "project.createVersion",
@@ -390,6 +394,8 @@ export const operations: Manifest = [
       { name: "maxResults", role: "query" },
       { name: "property", role: "query" },
     ],
+    // /user/search returns a bare array.
+    trim: "bareList",
   },
   {
     name: "user.get",
@@ -419,6 +425,8 @@ export const operations: Manifest = [
       { name: "actionDescriptorId", role: "query" },
       { name: "recommend", role: "query" },
     ],
+    // /user/assignable/search returns a bare array.
+    trim: "bareList",
   },
   {
     name: "user.bulkGet",
@@ -430,6 +438,7 @@ export const operations: Manifest = [
       { name: "startAt", role: "query" },
       { name: "maxResults", role: "query" },
     ],
+    trim: "list",
   },
 
   // =================================================================
@@ -448,6 +457,7 @@ export const operations: Manifest = [
       { name: "name", role: "query" },
       { name: "projectKeyOrId", role: "query" },
     ],
+    trim: "list",
   },
   {
     name: "board.get",
@@ -500,6 +510,7 @@ export const operations: Manifest = [
       { name: "startAt", role: "query" },
       { name: "maxResults", role: "query" },
     ],
+    trim: "list",
   },
   {
     name: "board.backlog",
@@ -514,6 +525,7 @@ export const operations: Manifest = [
       { name: "startAt", role: "query" },
       { name: "maxResults", role: "query" },
     ],
+    trim: "list",
   },
   {
     name: "board.epics",
@@ -527,6 +539,7 @@ export const operations: Manifest = [
       { name: "startAt", role: "query" },
       { name: "maxResults", role: "query" },
     ],
+    trim: "list",
   },
 
   // =================================================================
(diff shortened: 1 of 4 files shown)


src/core/operations.ts
```diff
@@ -976,6 +997,10 @@ export const operations: Manifest = [
       { name: "caseInsensitive", role: "query" },
       { name: "maxResults", role: "query" },
     ],
+    // /groups/picker returns { groups: [...], total }; treat as a
+    // paginated list — paginatedListSummary's pageItemCount falls
+    // through to the array length when `groups` isn't a known key.
+    trim: "list",
```
scottlepp, line 1003: The inline comment here is inaccurate: "`groups` isn't a known key" — but `groups` *is* explicitly listed in `pageItemCount`'s `??` chain (`trim.ts` line 223), so it will be picked up correctly. The fallback-to-array-length path that the comment describes (`?? 0` at the end of the chain) is only reached when *none* of the five known keys are present.

The comment should read something like: "`groups` is a recognized key in `pageItemCount`'s `??` chain, so `truncated` and the fallback totals will be set correctly from the `groups` array." The current wording will confuse anyone who later adds a new key and wonders whether `groups` was intentionally excluded.

tests/core/trim.test.ts
```diff
@@ -242,3 +246,114 @@ describe("searchSummary", () => {
...
+    };
+    const s = voteListSummary(r);
+    expect(s).toEqual({ votes: 3, hasVoted: false, truncated: true });
+  });
+
+  it("falls back to voters.length when votes is missing", () => {
+    const s = voteListSummary({ voters: [{}] });
+    expect(s.votes).toBe(1);
+  });
+});
```
scottlepp, line 367: `voteListSummary` is missing a null/undefined input test, unlike `watcherListSummary` which has one directly above. The implementation handles null correctly (`r?.voters ?? []` short-circuits), but this edge case is undocumented by the test suite. Suggested addition inside the `voteListSummary` describe block:

```ts
it("treats null/empty input as empty", () => {
  expect(voteListSummary(null)).toEqual({ votes: 0, hasVoted: undefined, truncated: false });
  expect(voteListSummary(undefined)).toEqual({ votes: 0, hasVoted: undefined, truncated: false });
});
```

scottlepp:
No blockers. 2 nit/suggestion comments posted inline. Addressing the five questions directly:

**1. Coverage gaps in the manifest sweep**

The user.search vs user.bulkGet split is correct. `/user/search` returns a bare array; `/user/bulk` returns a PageBean with a `values` array — `trim: "bareList"` and `trim: "list"` are the right calls respectively. The agile endpoints (board.list, sprint.listForBoard, board.epics) all return `{ values: [...], total, ... }`, which `pageItemCount` recognizes. No missed list-shape operations visible in the manifest.

**2. The shape choice — count-only vs search-shaped with inline rows**

The asymmetry is justified for board.issues/sprint.issues/epic.issues. The key reason searchSummary inlines rows is that an agent running a JQL search is typically selecting *which* issue to act on next — the key is the output. An agent calling board.issues or sprint.issues is typically asking "how many are there" or "do these exist" before deciding whether to paginate. Inlining issue rows from board.issues would reproduce searchSummary at the cost of a second code path with no clear benefit, since the agent has `issue.search` if they want to select from a board's backlog by JQL anyway. The count-only shape is the right default here.

**3. The `??` chain ordering in `pageItemCount`**

Not brittle for any real Jira endpoint. The five array keys (`comments`, `worklogs`, `issues`, `values`, `groups`) do not overlap in any current operation's response: `/issue/{key}/comment` returns `comments`, `/issue/{key}/worklog` returns `worklogs`, board/sprint/epic list endpoints return `issues`, `/filter/search` and Agile board/sprint list endpoints return `values`, and `/groups/picker` returns `groups`. No current endpoint returns two of these simultaneously. If a future endpoint did, the first match in the chain would win silently — worth a note in the `PageBean` type comment, but not a bug today.

One thing to fix: the comment on `group.search` in operations.ts says "`groups` isn't a known key" — it is recognized (line comment on that change). See inline comment.

**4. `watcherListSummary.isWatching` — undefined vs omitted**

Not a meaningful distinction for an LLM agent. JSON.stringify serializes `undefined` properties as omitted already (`JSON.stringify({ a: undefined })` → `"{}"`), so the wire representation is identical whether the field is `undefined` or absent. The agent never sees a literal `undefined`. The current behavior is fine.

**5. `searchSummary` as a justified exception**

50 rows × ~80B = ~4KB is acceptable under this PR's philosophy. The materiality threshold here is the 84KB→2.8KB reduction this PR achieves on list endpoints. searchSummary was already at ~4KB before this PR and is untouched — it's not the problem this PR is solving. The inline-rows rationale (agent picks a key, follows up) remains valid and the size is two orders of magnitude below the problematic cases. The exception is justified; no change needed.

src/core/operations.ts
scottlepp: Addressed in fac2ae1 — comment now describes the actual behavior: `groups` is a recognized key in `pageItemCount`'s `??` chain, so the fallback-to-zero path only fires when none of the five known keys are present.

tests/core/trim.test.ts
scottlepp: Addressed in fac2ae1 — `voteListSummary` now has a symmetric null-input test next to `watcherListSummary`'s.

PR #192, opened 2026-05-05
v2: flip default tool mode from code-api to classic

scottlepp:
## Summary
The post-PR #180 benchmark showed classic mode ties or beats code-api on every per-call scenario. Code-api's only remaining edge is the tool-list cost (~95 tokens vs ~7,800 for classic, paid once per session). That's a real saving but not worth the per-call complexity (`JIRA_MCP_SOCKET=… npx tsx -e '…'` on every Jira call) for the common case.

| | classic | code-api summary |
| --- | --- | --- |
| simple ticket | 1.1KB | 1.3KB |
| rich investigation | 2.8KB | 3.1KB |
| JQL search ~10 tickets | 3.4KB | 3.6KB |

Code-api stays available as `JIRA_TOOL_MODE=code-api` for sessions that do many composed queries (jq filters, multi-call investigations) where the tool-list saving compounds across many turns.

## Changes
- `parseToolMode("")` returns `"classic"` instead of `"code-api"`. Same for unset.
- Updated the comment on the `ToolMode` type to reflect the new default.
- Updated the env-var help text in the missing-config error message.
- Flipped two test expectations and adjusted test names accordingly.

## Out of scope (planned PR (c2))
Port v1's `JIRA_ENABLED_CATEGORIES` / `JIRA_DISABLED_TOOLS` env-var filtering to v2's consolidated tools, so users who want to cut the tool-list cost without giving up classic-mode simplicity can. With both this PR and (c2) merged, classic + filtering can get the tool list down to ~1.5KB while keeping the simple per-call shape.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 380 passed (no behavioral change in non-config tests)
- [x] Smoke test: `node build/index.js` (no env) boots `mode=classic`; `JIRA_TOOL_MODE=code-api node build/index.js` still boots in code-api mode

### DIFF
--- a/src/codeapi/tool.ts
+++ b/src/codeapi/tool.ts
@@ -1,6 +1,6 @@
 // The single MCP tool exposed in code-api mode (PR #10).
 //
-// In default mode the server publishes only this tool. Calling it
+// In code-api mode the server publishes only this tool. Calling it
 // returns the path to the generated TypeScript stubs and a short
 // usage example; the agent then drives Jira through those stubs in
 // its own execution environment (Claude Code's Bash + tsx, etc.) and
--- a/src/config.ts
+++ b/src/config.ts
@@ -12,12 +12,16 @@ config({ path: ".env.local", override: true });
 export type TokenType = "scoped" | "classic";
 
 // v2 tool surface modes:
-// - "code-api": (default) expose a single `jira_code_api` MCP tool that
-//   points at on-disk TypeScript stubs the agent imports via tsx. Most
-//   token-efficient — ~500 tokens for the tool listing.
-// - "classic": expose the 16 consolidated MCP tools. Keeps the
-//   request/response shape v1 users are used to. ~3-4k tokens for the
-//   tool listing.
+// - "classic": (default) expose the 16 consolidated MCP tools. Same
+//   request/response shape v1 users are used to, with v2's trim/sandbox
+//   on the response side. The default for v2.0 because per-call cost is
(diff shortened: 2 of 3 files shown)


scottlepp:
One nit, no blockers.

**Stale comment in an untouched file — `src/codeapi/tool.ts`, line 3**

The file-level block comment reads:

> "In **default mode** the server publishes only this tool."

After this PR, code-api is opt-in, not the default. The sentence should be:

> "In **code-api mode** the server publishes only this tool."

This file wasn't changed in the diff so GitHub won't let me pin a line comment to it, but it's the one place that still says "default mode" in a way that refers to code-api being the default.

---

Everything else the author flagged is clean:

- `src/index.ts` dispatches on `cfg.toolMode` dynamically — no hardcoded default.
- `JIRA_CODE_API_TOOL_DESCRIPTION` does not mention "default".
- `src/codeapi/boot.ts` and `generator.ts` have no stale default claims (the `defaultRefsImportPath` function name is unrelated to tool mode).
- All three edits in `src/config.ts` (return value, error message, help text) are mutually consistent.
- The test file deletes `JIRA_TOOL_MODE` in `beforeEach`, so dotenv / `.env.local` leakage cannot affect the unset-default assertion — that test is real.
- The existing "throws on unknown mode" test covers the error message; an additional test for the word order of "classic" vs "code-api" in the message would test human-readable help text, not a contract, so I wouldn't add it.

PR #193, opened 2026-05-05
v2: env-var tool filtering (PR (c2))

scottlepp:
## Summary
Restores v1/main's tool-list trimming via env vars, adapted to v2's consolidated-tool surface. Two independent knobs:

```
JIRA_ENABLED_CATEGORIES=issue,search,comment
JIRA_DISABLED_ACTIONS=issue.delete,project.delete
```

Both optional. Empty = all enabled.

## Tool-list cost reduction (classic mode)

| filter | bytes | ~tokens | factor |
| --- | --- | --- | --- |
| none (16 tools) | 30.6KB | ~7,800 | 1× |
| 3 categories | 6.3KB | ~1,600 | 5× |
| 3 cats + 5 disabled actions | 4.6KB | ~1,200 | 6.6× |

For users who don't use the full Jira surface, classic + filter now lands well under 2KB tool-list cost while keeping classic's simpler per-call shape. Code-api mode is still ~95 tokens for users who want maximum compression at the cost of `JIRA_MCP_SOCKET=… npx tsx -e '…'` per call.

## Safety

The disabled-action check is enforced at the manifest dispatch layer (`invokeOperationRaw`), so it applies to **both** modes. A destructive op disabled in classic stays disabled if a user later opts into code-api — the bridge's `invokeAndSandbox` rejects it before any HTTP call.

## Category-name caveat
Categories use the consolidated tool name (the part after `jira_`), not the manifest's category prefix. Two of those differ:
- `issueLink.*` operations → `jira_link` tool → category name `link`
- `permissions.*` and `vote.*` don't have standalone tools (they live under `jira_group` and `jira_watcher` actions); their ops can still be disabled via `JIRA_DISABLED_ACTIONS=permissions.mine,vote.add`.

Documented in `src/config.ts`.

## Implementation
- **`src/config.ts`** — adds `ToolFilterConfig` (`enabledCategories`, `disabledActions`). `parseToolFilter` validates categories against a canonical list and warns on typos rather than throwing. Action names go through unvalidated; the manifest layer rejects unknown ones at first use.
- **`src/core/manifest.ts`** — new `assertOperationEnabled()` helper called from `invokeOperationRaw`. Throws `OperationError` with a `JIRA_DISABLED_ACTIONS` message before any HTTP call.
- **`src/codeapi/bridge.ts` + `boot.ts`** — `disabledActions` plumbed through `StartBridgeOpts` → `handleConnection` → `dispatchLine` → `invokeAndSandbox`.
- **`src/tools/v2/index.ts`** — replaces static `v2Tools` with `getV2Tools(filter)`. Drops categories not in the whitelist and trims disabled actions out of each remaining tool's `oneOf`. Tools with every action disabled are dropped entirely. `v2Tools` still exported for back-compat.
- **`src/tools/v2/dispatcher.ts`** — `disabledActions` threaded through `dispatchTool` → `invokeOperation`.
- **`src/index.ts`** — captures the parsed config at boot, passes the filter to `getV2Tools` (classic) or `bootCodeApi` (code-api), and surfaces the active filter on stderr at startup.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 416 passed (+36 new), 1 skipped
- [x] Smoke test: `JIRA_ENABLED_CATEGORIES=issue,search JIRA_DISABLED_ACTIONS=issue.delete node build/index.js` boots `mode=classic, tools=2` with the filter logged to stderr; unfiltered run boots `tools=16`.

### DIFF
--- a/src/tools/v2/index.ts
+++ b/src/tools/v2/index.ts
@@ -80,7 +80,12 @@ function categoryOf(toolName: string): string {
 // emits the full 16-tool surface (back-compat with how the MCP server
 // called this before c2). With a filter:
 //   - Tools whose category isn't in `enabledCategories` are dropped
-//     entirely. Empty `enabledCategories` means "all enabled".
+//     from the listing. Empty `enabledCategories` means "all enabled".
+//     This is a token-cost knob, not a hard enforcement: the dispatch
+//     path (`handleV2Tool`) doesn't consult `enabledCategories`, so a
+//     misbehaving or stale-cache agent could still reach a hidden
+//     tool. Use `disabledActions` for the safety guarantee — it's
+//     enforced at the manifest dispatch layer in both modes.
 //   - Actions in `disabledActions` are stripped from each tool's
 //     `oneOf`, so the disabled action's schema doesn't bloat the
 //     tool listing. A tool with every action disabled is dropped.
--- a/src/tools/v2/index.ts
+++ b/src/tools/v2/index.ts
@@ -1,6 +1,8 @@
 // Public surface for v2 consolidated tools.
 //
-// `v2Tools` is the MCP-shaped tool listing (handed to ListToolsRequest);
+// `getV2Tools(filter)` builds the MCP-shaped tool listing (handed to
+// ListToolsRequest), optionally honoring a ToolFilterConfig that
+// drops categories and disables individual actions.
 // `handleV2Tool` is the call dispatcher (called from
 // CallToolRequestSchema's handler in src/index.ts).
 //
@@ -9,10 +11,12 @@
 // against the manifest in src/core/operations.ts.
 
 import type { JiraClient } from "../../auth/jira-client.js";
+import type { ToolFilterConfig } from "../../config.js";
 import { operations } from "../../core/operations.js";
 import {
   buildInputSchema,
   dispatchTool,
+  type ActionDefinition,
   type ConsolidatedTool,
 } from "./dispatcher.js";
 import { jiraAttachment } from "./attachment.js";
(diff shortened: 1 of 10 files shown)


tests/tools/v2/filter.test.ts
```diff
@@ -0,0 +1,170 @@
...
+import { describe, expect, it } from "vitest";
+
+import type { ToolFilterConfig } from "../../../src/config.js";
+import {
+  allConsolidatedTools,
+  getV2Tools,
+  v2Tools,
+} from "../../../src/tools/v2/index.js";
+
+const ALL_CATEGORIES = [
```
scottlepp: The `ALL_CATEGORIES` constant here is a hardcoded duplicate of `CONSOLIDATED_CATEGORIES` in `src/config.ts`. If someone adds a 17th tool and updates `CONSOLIDATED_CATEGORIES` (the config-side canonical list), they also have to remember to update this array — but nothing enforces that. The round-trip test at line 849 (`it.each(ALL_CATEGORIES)(...)`) won't cover the new category, so the drift goes undetected in exactly the scenario the test was designed to catch.

Suggestion: replace the inline literal with an import of `CONSOLIDATED_CATEGORIES` from `../../../src/config.js`. The test already imports from config via `ToolFilterConfig`, so there's no new dependency direction.

src/tools/v2/index.ts
```diff
@@ -64,11 +68,57 @@ export interface V2Tool {
...
+// src/tools/v2/link.ts).
+function categoryOf(toolName: string): string {
+  return toolName.startsWith("jira_") ? toolName.slice(5) : toolName;
+}
+
+// Build the MCP tool listing for classic mode. With no filter passed,
+// emits the full 16-tool surface (back-compat with how the MCP server
+// called this before c2). With a filter:
+//   - Tools whose category isn't in `enabledCategories` are dropped
+//     entirely. Empty `enabledCategories` means "all enabled".
```
scottlepp: The `handleV2Tool` dispatcher uses `toolByName`, which is built from all 16 tools unconditionally (line 58-59). This means a caller that passes a tool name for a category that was filtered out of `getV2Tools` will still get a successful dispatch. The category filter is listing-only, not call-time.

The PR description and the comment on `ToolFilterConfig` both explain the asymmetry clearly (categories = listing filter, disabledActions = safety guarantee). However, the comment on `getV2Tools` says 'Tools whose category isn't in `enabledCategories` are dropped entirely', which could be read as applying at dispatch time too. A reader landing on this function without the PR context would not expect a 'dropped' tool to still be callable.

Not a bug given the stated design, but the comment is subtly misleading. Suggest adding a parenthetical: '... dropped from the MCP listing (but still dispatchable — use `disabledActions` for hard enforcement)'.

tests/tools/v2/filter.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    const out = getV2Tools({
+      enabledCategories: ["issue"],
+      disabledActions: ["issue.delete", "issue.create"],
+    });
+    expect(out.map((t) => t.name)).toEqual(["jira_issue"]);
+    const actions = actionsIn(out[0]);
+    expect(actions).not.toContain("delete");
+    expect(actions).not.toContain("create");
+    expect(actions).toContain("get");
+  });
```
scottlepp, line 140: The 'compose' test has `enabledCategories: ["issue"]` with actions disabled from the same category. You asked whether to add a test for the cross-category case: e.g. `enabledCategories: ["search"]` with `disabledActions: ["issue.delete"]`.

The current implementation handles it correctly — the category check skips `jira_issue` before the action filter runs, so `issue.delete` in `disabledActions` is irrelevant for the listing (though still enforced at dispatch time by `assertOperationEnabled`). That ordering is implicit and easy to get wrong in a future refactor.

Suggest adding one case that puts a disabled action in a category that is also excluded, and asserts both that the excluded tool is absent and that the retained tool is unaffected. This is a low-cost test that pins the 'category filter runs before action filter' invariant explicitly.

scottlepp:
No blockers. Three suggestions, no correctness bugs found.

**Addressing your specific questions:**

1. **Category name surface mismatch** — The warning + comment is fine. The fact that a valid `issueLink` in v1 silently produces zero tools is already covered by the test at filter.test.ts:96. The warning message could mention 'use the tool name suffix (after `jira_`), not the manifest category prefix' to preempt the v1 footgun, but this is a nit.

2. **`CONSOLIDATED_CATEGORIES` drift risk** — This is the real gap. The `ALL_CATEGORIES` constant in `filter.test.ts` is a hardcoded duplicate of `CONSOLIDATED_CATEGORIES` in `config.ts`. The `it.each(ALL_CATEGORIES)` sanity test only covers categories listed in the test file, so if a 17th tool is added and `CONSOLIDATED_CATEGORIES` is updated but `ALL_CATEGORIES` is not, the new category is silently untested. Fix: import `CONSOLIDATED_CATEGORIES` directly in the test. (See line comment on `tests/tools/v2/filter.test.ts:17`.)

3. **Dual-layer disabled-action filter** — The duplication is necessary. Code-api mode bypasses `getV2Tools` entirely (the listing is a single `jira_code_api` tool, not the 16 consolidated tools), so the listing-layer filter wouldn't apply. The manifest-layer `assertOperationEnabled` check in `invokeOperationRaw` is the only enforcement path for code-api. Keeping it there is the right call.

4. **Filter matrix edge case** — The cross-category case (`enabledCategories: ["search"]` + `disabledActions: ["issue.delete"]`) works correctly: the category check skips `jira_issue` before the action filter runs. The behavior is correct but the ordering invariant is untested. See line comment on filter.test.ts:156.

5. **Positional 5th arg** — Acceptable here. The function was already at 4 positional params, all of them manifest/client/name/args that are conceptually different from the filter. A 5th optional that's semantically distinct doesn't create ambiguity the way two adjacent strings would. No change needed.

6. **Dropped path check** — Wired through all 5 call sites: `index.ts → handleV2Tool`, `dispatcher.ts → dispatchTool → invokeOperation`, `bridge.ts → dispatchLine → invokeAndSandbox → invokeOperationRaw`, `boot.ts → startBridge`. No dropped paths found.

**One additional nit (line comment on `src/tools/v2/index.ts:83`):** The `getV2Tools` comment says 'Tools whose category isn't in `enabledCategories` are dropped entirely', which could be read as meaning they're also blocked at dispatch time. `toolByName` still covers all 16 tools, so a caller that names a category-filtered tool will succeed at dispatch. The PR description explains this asymmetry clearly (categories = listing filter, disabledActions = safety guarantee), but the in-code comment doesn't. Minor wording fix would help future readers.

PR #194, opened 2026-05-05
v2: docs, migration guide, integration tests (PR #11)

scottlepp:
## Summary
Bundles the three remaining pieces from the plan's PR #11:
- **README rewrite** for v2's tool surface, env vars, and code-api opt-in flow.
- **`docs/MIGRATION.md`** — full v1→v2 tool-name mapping for every v1 surface, env-var renames, response shape change, code-api section.
- **`CHANGELOG.md`** — v2.0 entry with the published benchmark numbers from PR #179/#180.
- **`tests/integration.test.ts`** — replaces the PR #7c placeholder skip with a real live-Jira suite covering both classic and code-api modes. Read-only by design. Skips when `JIRA_HOST` is unset so CI without secrets keeps passing.

## Bonus fixes the integration suite surfaced

- **`searchSummary` (src/core/trim.ts):** Jira's `GET /search/jql` v3 dropped `total`/`startAt`/`maxResults` in favor of cursor-based `isLast`/`nextPageToken`. The trim made `total` required, so the integration test crashed on a real response. Made the legacy fields optional, added `isLast`/`nextPageToken` passthrough, guarded against missing `issues` array. Two regression unit tests added.
- **`package.json`:** adds `npm run test:integration`, bumps `engines.node` to `>=20` (matches the plan's PR #1 spec).

## Numbers

429 tests pass total (was 417 + 1 skipped). New:
- 10 live-Jira integration tests covering classic + code-api end-to-end.
- 2 unit tests locking in `searchSummary`'s new optional shape.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 429 pass, 0 skipped (with creds in `.env.local`)
- [x] Verified the integration suite skips cleanly with no `.env*` files (`describe.skipIf(!canRun)` covers it)
- [x] Manual smoke: README's claude_desktop_config.json snippet still works (same shape as before, new env vars optional)

## Out of scope
PR #12 — bumping `package.json` to `2.0.0`, tagging the release, publishing to npm. The CHANGELOG entry is in place; just needs the version bump and `git tag`.

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -1,0 +1,263 @@
+# Migrating from jira-mcp v1 to v2
+
+v2 is a clean break, not a drop-in. Tools have new names, the response shape changed, and one env var got renamed. Most v1 users will spend a few minutes updating their MCP server config and then never need to think about it again. v1 is still maintained on the [`1.x` branch](https://github.com/scottlepp/jira-mcp/tree/1.x) if you'd rather stay there.
+
+## What changed at a glance
+
+- **85 tools → 16 consolidated tools.** Each v1 tool name like `jira_get_issue` becomes a `(tool, action)` pair: `jira_issue` with `action: "get"`. Tool-list cost drops from ~7,800 tokens to ~1,600–4,600 depending on how aggressively you filter (and to ~95 in the optional `code-api` mode).
+- **Responses are now `{summary, ref, ...}` instead of raw Jira JSON.** The summary is a trimmed projection inline; the full untrimmed response is written to `${TMPDIR}/jira-mcp/${session}/` and the `ref` field points at it. Read the ref when you need detail. v1 returned the entire 50–100KB Jira payload.
+- **`JIRA_DISABLED_TOOLS` is now `JIRA_DISABLED_ACTIONS`** and takes manifest operation names like `issue.delete` instead of v1 tool names like `jira_delete_issue`.
+- **`JIRA_ENABLED_CATEGORIES` keeps working** with one rename: `issueLink` is now `link` (matches the v2 tool's name `jira_link`).
+- **New optional env var: `JIRA_TOOL_MODE`** (`"classic"` default, `"code-api"` opt-in for power users). Most users should leave this unset.
+
+## What you need to change
+
+### Step 1: Tool calls
+
+If you have any code, prompts, or stored sessions that name a v1 tool directly, update to the v2 shape.
+
+**v1:**
+```json
+{ "name": "jira_get_issue", "arguments": { "issueIdOrKey": "PROJ-1" } }
+```
+
+**v2:**
+```json
+{ "name": "jira_issue", "arguments": { "action": "get", "issueIdOrKey": "PROJ-1" } }
+```
+
+The full mapping is at the bottom of this file. For most users this is a one-time edit if you've baked tool names into anything. For agents driving Jira through Claude, the agent reads the tool listing fresh each session and adapts automatically.
+
+### Step 2: Update env vars
+
+If you used `JIRA_DISABLED_TOOLS`, switch to `JIRA_DISABLED_ACTIONS` and translate the values:
+
+| v1 (`JIRA_DISABLED_TOOLS`) | v2 (`JIRA_DISABLED_ACTIONS`) |
+|---|---|
+| `jira_delete_issue` | `issue.delete` |
+| `jira_delete_project` | `project.delete` |
+| `jira_delete_comment` | `comment.delete` |
+| `jira_delete_attachment` | `attachment.delete` |
+| `jira_delete_worklog` | `worklog.delete` |
+
+The format is `category.action` — same as v2's manifest operation names.
+
+If you used `JIRA_ENABLED_CATEGORIES` with `issueLink`, change it to `link`. All other category names are unchanged.
+
+### Step 3: Adjust to the new response shape (only if you wrote code that consumes the JSON)
+
+For agents using Claude/Claude Code: nothing to do — Claude reads the new shape and adapts. For programmatic consumers:
+
+**v1 response (raw Jira):**
+```json
+{ "id": "10001", "key": "PROJ-1", "fields": { "summary": "...", "status": {...}, "comment": { "comments": [...] }, ... } }
+```
+
+**v2 response (sandboxed):**
+```json
+{
+  "summary": { "key": "PROJ-1", "status": "In Progress", "assignee": {...}, "recentComments": [...] },
+  "ref": "/tmp/jira-mcp/abc123/issue-get/<hash>.json",
+  "hash": "abc123def456...",
+  "fullSize": 47823,
+  "fetchedAt": "2026-05-05T12:34:56Z"
+}
+```
+
+Read `ref` (it's an absolute path to the full untrimmed Jira response on disk) only when you need fields the summary omits.
+
+## Full tool name mapping
+
+Source of truth: every v1 tool name in v1's README, mapped to its v2 equivalent.
+
+### Issues → `jira_issue`
+
+| v1 | v2 action |
+|---|---|
+| `jira_get_issue` | `get` |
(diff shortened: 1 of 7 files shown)


docs/MIGRATION.md
```diff
@@ -0,0 +1,263 @@
...
+**v1 response (raw Jira):**
+```json
+{ "id": "10001", "key": "PROJ-1", "fields": { "summary": "...", "status": {...}, "comment": { "comments": [...] }, ... } }
+```
+
+**v2 response (sandboxed):**
+```json
+{
+  "summary": { "key": "PROJ-1", "status": "In Progress", "assignee": {...}, "recentComments": [...] },
+  "ref": "/tmp/jira-mcp/abc123/issue-get/<hash>.json",
```
scottlepp, line 89: The v2 response shape shown here is the **code-api / bridge** shape (`SandboxResult`), not what classic tools return.

In classic mode, `handleV2Tool` calls `invokeOperation`, which applies a trim projection and returns it directly — no `sandbox()` call, no disk write, no `ref` field. A real classic response for `jira_issue { action: "get" }` looks like:

```json
{
  "key": "PROJ-1",
  "id": "10001",
  "summary": "...",
  "status": "In Progress",
  "assignee": {...},
  "commentCount": 4,
  "recentComments": [...],
  "attachmentCount": 0,
  "attachments": []
}
```

The `{ "summary": {...}, "ref": "...", "hash": "...", "fullSize": ..., "fetchedAt": "..." }` envelope only appears when using code-api mode (bridge path via `invokeOperationRaw + sandbox()`). The migration guide's step 3 as written will mislead any programmatic consumer who is migrating from classic v1 to classic v2.

docs/MIGRATION.md
```diff
@@ -0,0 +1,263 @@
...
+
+| v1 (`JIRA_DISABLED_TOOLS`) | v2 (`JIRA_DISABLED_ACTIONS`) |
+|---|---|
+| `jira_delete_issue` | `issue.delete` |
+| `jira_delete_project` | `project.delete` |
+| `jira_delete_comment` | `comment.delete` |
+| `jira_delete_attachment` | `attachment.delete` |
+| `jira_delete_worklog` | `worklog.delete` |
+
+The format is `category.action` — same as v2's manifest operation names.
```
scottlepp: The sentence "The format is `category.action` — same as v2's manifest operation names" is correct in general, but creates a trap for the `jira_link` category. The `jira_link` tool's manifest operation names use the prefix `issueLink`, not `link`:

```
jira_link action "create" → manifest operation "issueLink.create"
jira_link action "delete" → manifest operation "issueLink.delete"
```

So a user who wants to disable link creation after reading this table will naturally try `link.create` or `link.delete`, which will silently do nothing — `assertOperationEnabled` in `manifest.ts` checks `disabledActions.includes(name)` and the name never matches. No warning is emitted for unrecognised entries.

The fix is either to add a row to the table (e.g. `jira_delete_issue_link → issueLink.delete`) or add a note alongside the step 2 `JIRA_ENABLED_CATEGORIES` rename that the corresponding `JIRA_DISABLED_ACTIONS` names for link ops keep the `issueLink.` prefix.

src/core/trim.ts
```diff
@@ -141,18 +141,37 @@ export const issueSummary = (i: JiraIssue): IssueSummary => {
...
-  startAt: number;
-  maxResults: number;
+  // Known when the server returns the legacy PageBean shape; absent
+  // when /search/jql responds. Callers should treat these as hints,
+  // not hard counts.
+  total?: number;
+  startAt?: number;
+  maxResults?: number;
+  // Cursor-based pagination signal from /search/jql. `false` (or
+  // missing `nextPageToken`) means more pages are available.
```
scottlepp: The comment has the semantics of `nextPageToken` backwards. "missing `nextPageToken`` means more pages are available" is wrong — a missing (absent) `nextPageToken` means you are **on the last page** (no cursor = nowhere to continue). The correct reading is:

- `isLast: false` → more pages available (comment has this right)
- `nextPageToken` **present** → a next page exists; follow the cursor
- `nextPageToken` **absent** → last page reached

Suggested replacement:
```ts
// Cursor-based pagination signal from /search/jql. 
// means more pages are available. When present,  is
// the cursor to pass on the next request; its absence signals the
// last page has been reached.
```

The runtime code is fine (it just passes the fields through); only the comment misleads.

scottlepp:
3 line comments posted (2 blockers, 1 nit). No other issues found.

**Blockers:**

- **docs/MIGRATION.md line 60** — Step 3's "v2 response (sandboxed)" example shows the code-api/bridge `SandboxResult` envelope (`{summary:{...}, ref:"...", hash:"...", fullSize:..., fetchedAt:"..."}`), but classic tools do not go through `sandbox()`. Classic mode returns the trim projection directly (e.g. `IssueSummary` with `key`, `status`, `commentCount` at the top level, no `ref` field). Verified: `invokeOperation` in `manifest.ts` applies the trim projection and returns it without calling `sandbox()`; the `src/index.ts` handler passes the result straight to `textResponse()`. The `{summary,ref}` envelope is exclusive to the bridge path. A programmatic v1→v2 migrant consuming classic tools will look for `result.summary.key` instead of `result.key` and break. The integration test itself correctly accesses `result.key` (not `result.summary.key`), confirming the discrepancy.

- **docs/MIGRATION.md line 43** — Step 2 says "The format is `category.action` — same as v2's manifest operation names", which is correct for every category except `link`. The `jira_link` tool's manifest operation names carry the prefix `issueLink` (e.g. `issueLink.create`, `issueLink.delete`), not `link`. A user who sets `JIRA_DISABLED_ACTIONS=link.delete` gets no error — `assertOperationEnabled` does a literal string match and the unrecognised entry is silently ignored — so the link operation remains callable. Needs either a row in the step 2 table or an explicit note that link ops use `issueLink.` as their action prefix.

**Nit:**

- **src/core/trim.ts line 157** — The comment says "missing `nextPageToken`) means more pages are available", which is backwards. An absent `nextPageToken` means the last page has been reached (no cursor to follow). Runtime behavior is unaffected since the field is just passed through; only the comment is wrong.

---

**Responses to the author's specific questions:**

1. **MIGRATION.md correctness (Issue, Comment, Issue Links sections):** The tool-action mapping table is correct against the source — verified by diffing every action key in the 16 `.ts` tool files. The Issue Links note about "renamed from issueLink to jira_link" is accurate, but the step 2 env-var guidance creates the silent trap described above.

2. **`searchSummary` shape change and CHANGELOG:** Making the three fields optional is the right call. It is technically a narrowing of the `SearchSummary` type contract, but since v2 hasn't shipped and the only callers are the integration test (loose cast) and agents (JSON), flagging it in CHANGELOG as a breaking internal change would add noise without value. Leave it alone.

3. **CI safety / partial env:** The `canRun` AND-gate requires all five vars (`JIRA_HOST`, `JIRA_EMAIL`, `JIRA_API_TOKEN`, `JIRA_BENCH_TICKET_RICH`, `JIRA_BENCH_TICKET_SIMPLE`). `describe.skipIf(!canRun)` applies to both `describe` blocks including their `beforeAll` hooks — vitest skips lifecycle hooks inside a skipped describe. Partial-env (JIRA_HOST set but token missing) is safe because `haveTickets` will be false on any CI that doesn't set the bench vars, keeping `canRun` false. No gap there.

4. **Documentation accuracy / README vs MIGRATION cross-check:** The 16-tool action tables in README.md and MIGRATION.md match each other and match the source definitions.

5. **Integration suite shape:** Read-only coverage is appropriate for a live-Jira suite. The mock-client unit tests in `tests/tools/v2/tools.test.ts` already exercise create/update/transition handlers. Adding live mutation tests risks leaving debris and is out of the stated PR scope.

6. **The `callBridge` TCP address parser:** The `tcp:127.0.0.1:<port>` format emitted by the bridge on Windows is parsed correctly — `lastIndexOf(":")` finds the port separator, `slice(4, lc)` extracts `127.0.0.1`. No bug there.

PR #195, opened 2026-05-05
v2: flatten consolidated tool input schemas

scottlepp:
## Summary
- The Anthropic tool-use API rejects top-level `oneOf`/`allOf`/`anyOf` in `input_schema`, so the v2 consolidated tools were unusable in classic mode (`API Error: 400 ... input_schema does not support oneOf, allOf, or anyOf at the top level`).
- Replaces the per-action `oneOf` branches in `buildInputSchema` with a flat object schema: `action` is a string enum, fields are merged across actions, per-action required/optional info moves into the tool description.
- Runtime Zod validation in `dispatchTool` still enforces per-action arg shape, so malformed calls fail fast — just after the call instead of at schema validation.

## Why this also reduces tokens
The old shape duplicated every shared field across each action's branch (e.g. `issueIdOrKey` appeared in 6 of `jira_issue`'s 9 branches). The new shape lists each field once. Net effect: smaller tool listing for any tool with overlapping fields across actions.

## Trade-off
The JSON Schema no longer encodes "field X is required only when action=Y". The agent gets that signal from the description (`- get: Fetch a single issue. (requires issueIdOrKey; optional fields, expand)`) instead of the schema. Code-api mode is unchanged.

## Test plan
- [x] `npx vitest run` — 430/430 passing
- [x] `npx tsc --noEmit` — clean
- [ ] Manual: point a Claude client at this branch and confirm the tool listing is accepted (no 400)

### DIFF
--- a/src/tools/v2/dispatcher.ts
+++ b/src/tools/v2/dispatcher.ts
@@ -69,7 +69,14 @@ export interface ConsolidatedTool {
 // control over the wire shape, (b) it'd add a dep, and (c) the action
 // schemas are small enough that hand-shaped JSON is clearer.
 export function buildInputSchema(tool: ConsolidatedTool): unknown {
-  const merged: Record<string, unknown> = {};
+  // Per-field collected JSON Schemas keyed by field name. We collect
+  // every action's contribution so that fields whose type differs
+  // across actions (e.g. jira_issue.fields is a CSV string for `get`
+  // but a record for `create`) are surfaced as a oneOf at the
+  // property level instead of silently first-wins. Top-level oneOf
+  // is what the Anthropic API rejects; nested oneOf under a property
+  // is fine.
+  const collected: Record<string, unknown[]> = {};
   const actionNames: string[] = [];
   const perActionLines: string[] = [];
 
@@ -79,7 +86,24 @@ export function buildInputSchema(tool: ConsolidatedTool): unknown {
     const required: string[] = [];
     const optional: string[] = [];
     for (const [key, sub] of Object.entries(shape)) {
-      if (!(key in merged)) merged[key] = zodFieldToJsonSchema(sub);
+      // Skip any per-action field literally named `action` — it
+      // would collide with the discriminator we add below. The
+      // dispatcher strips `action` from rawArgs before Zod
+      // validation, so an action schema declaring its own `action`
+      // field would be unreachable anyway. (jira_project.list has
+      // an `action` query param; this is the path that lets it
+      // coexist with the discriminator. The list action loses
+      // visibility of that param in the JSON Schema; agents can
+      // still pass it, additionalProperties:false is what blocks
+      // it — see below.)
+      if (key === "action") continue;
+      const js = zodFieldToJsonSchema(sub);
+      const seen = collected[key];
+      if (!seen) {
+        collected[key] = [js];
+      } else if (!seen.some((s) => deepEqual(s, js))) {
+        seen.push(js);
+      }
       (isOptionalZod(sub) ? optional : required).push(key);
     }
     const parts: string[] = [];
@@ -89,18 +113,50 @@ export function buildInputSchema(tool: ConsolidatedTool): unknown {
     perActionLines.push(`- ${actionName}: ${def.description}${suffix}`);
   }
 
+  const merged: Record<string, unknown> = {};
+  for (const [key, variants] of Object.entries(collected)) {
+    merged[key] = variants.length === 1 ? variants[0] : { oneOf: variants };
+  }
+
   const description = [tool.description, "Actions:", ...perActionLines].join("\n");
 
+  // Spread `merged` first so a (defensively-skipped) collision can
+  // never clobber the discriminator. additionalProperties is left
+  // permissive: per-action Zod validation in dispatchTool is the
+  // authoritative gate, and a strict top-level schema would reject
+  // legitimate per-action fields the merge couldn't represent
+  // (e.g. the `action` query param on jira_project.list).
   return {
     type: "object",
     description,
     properties: {
-      action: { type: "string", enum: actionNames },
       ...merged,
+      action: { type: "string", enum: actionNames },
     },
     required: ["action"],
-    additionalProperties: false,
+    additionalProperties: true,
   };
+}
+
+// Structural equality for the small JSON Schema fragments produced
+// by zodFieldToJsonSchema. Used to dedupe collected variants per
+// field — two actions with the same shape contribute one entry, two
+// with divergent shapes contribute a oneOf.
+function deepEqual(a: unknown, b: unknown): boolean {
+  if (a === b) return true;
+  if (typeof a !== typeof b || a === null || b === null) return false;
+  if (typeof a !== "object") return false;
+  if (Array.isArray(a)) {
+    if (!Array.isArray(b) || a.length !== b.length) return false;
+    return a.every((v, i) => deepEqual(v, b[i]));
+  }
+  if (Array.isArray(b)) return false;
+  const ao = a as Record<string, unknown>;
+  const bo = b as Record<string, unknown>;
+  const ak = Object.keys(ao);
+  const bk = Object.keys(bo);
+  if (ak.length !== bk.length) return false;
+  return ak.every((k) => deepEqual(ao[k], bo[k]));
 }
 
 // Reflect over a Zod schema using its constructor names + _def
--- a/src/tools/v2/dispatcher.ts
+++ b/src/tools/v2/dispatcher.ts
@@ -47,28 +47,59 @@ export interface ConsolidatedTool {
 
 // --- MCP-shaped JSON schema for the tool listing ----------------------
 
-// Build the input schema the MCP tool listing exposes. Uses oneOf so
-// the agent's tool-use renderer constrains output per action.
+// Build the input schema the MCP tool listing exposes.
+//
+// Shape: a flat object schema with `action` as a string enum and a
+// merged property bag containing every field across all actions. The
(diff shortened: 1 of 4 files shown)


src/tools/v2/dispatcher.ts
```diff
@@ -47,28 +47,59 @@ export interface ConsolidatedTool {
...
-      properties: zodToProperties(def.schema, actionName),
-      required: zodRequired(def.schema, actionName),
-      additionalProperties: false,
-    });
+    actionNames.push(actionName);
+    const shape = zodShape(def.schema) ?? {};
+    const required: string[] = [];
+    const optional: string[] = [];
+    for (const [key, sub] of Object.entries(shape)) {
+      if (!(key in merged)) merged[key] = zodFieldToJsonSchema(sub);
```
scottlepp: **Blocker: field-type collision silently produces the wrong JSON Schema for `jira_issue`**

`buildInputSchema` merges fields across all actions with a first-wins strategy (`if (!(key in merged))`). In `jira_issue`, the `get` action is declared first in `actions`, so its `fields: z.string()` wins — but `create`, `update`, and `transition` all expect `fields: z.record(...)`. The emitted schema for `jira_issue.fields` is therefore `{ type: "string" }`, which is the wrong type for three of the four actions that use the field.

Runtime Zod validation will still reject a bad call, but the agent will have been given a false type hint and may fail to construct a valid `create` call at all (since `fields` appears required for that action and the schema says it must be a string).

Concrete fix: name collisions with divergent types should be detected and merged as `oneOf` (or the narrower-type field should be renamed). Alternatively, add a cross-action duplicate-field check to the tool definitions and a test for the real `jira_issue` schema (not just `fixtureTool`).

src/tools/v2/dispatcher.ts
```diff
@@ -47,28 +47,59 @@ export interface ConsolidatedTool {
...
+  const description = [tool.description, "Actions:", ...perActionLines].join("\n");
+
   return {
     type: "object",
-    description: tool.description,
-    oneOf,
+    description,
+    properties: {
+      action: { type: "string", enum: actionNames },
+      ...merged,
```
scottlepp, line 133: **Blocker: `jira_project` `action` field overwrites the discriminator enum**

`project.ts` `ListSchema` declares `action: z.string().optional()` — a Jira API filter parameter meaning "what the caller is permitted to do with a project". When `buildInputSchema` runs for `jira_project`, `list` is the first action iterated and `action` lands in `merged` as `{ type: "string" }`.

The final properties object is assembled as:

```
properties: {
  action: { type: "string", enum: ["list","get","create",...] },  // hardcoded
  ...merged,  // contains  action: { type: "string" }  <-- overwrites
}
```

JavaScript spread evaluates left-to-right, so `merged`'s `action` key clobbers the `enum` discriminator. The emitted schema for `jira_project` will have `action: { type: "string" }` with no `enum` constraint — the agent cannot determine valid action names from the JSON Schema alone.

Fix: skip `"action"` when populating `merged` (it is always reserved for the discriminator). `jira_project.list`'s `action` Zod field would also need to be renamed (e.g. `permittedAction`) so it is not silently dropped from the schema listing.

scottlepp:
**2 blockers found. The core fix (removing top-level `oneOf`) is correct and the test coverage for that specific regression is solid, but the merge logic introduces two new bugs in the real tool definitions.**

---

**Blocker 1 — field-type collision in `jira_issue` (`dispatcher.ts` line 82)**

`buildInputSchema` uses first-wins merging (`if (!(key in merged))`). In `jira_issue` the `get` action is listed first, so its `fields: z.string()` registers before the `create`, `update`, and `transition` actions' `fields: z.record(...)`. The emitted JSON Schema advertises `fields` as `type: "string"` for all four actions. An agent trying to call `jira_issue` with `action: "create"` will see that type hint, send a string, and then hit a runtime Zod rejection — the exact trial-and-error the description says is avoided. The Jira issue `create` action will be effectively broken for agents that respect the schema.

Fix options: detect divergent types during merge and emit a `oneOf` for the field; or rename the `get`-action field (`fieldKeys` / `fieldSelector`) to break the collision.

**Blocker 2 — `action` discriminator clobbered for `jira_project` (`dispatcher.ts` line 99)**

`project.ts` `ListSchema` has `action: z.string().optional()` (the Jira API's project-permission filter). When `buildInputSchema` iterates `list` first, it writes `action: { type: "string" }` into `merged`. The hardcoded discriminator is then assembled as `{ action: { type: "string", enum: [...] }, ...merged }`. Because JavaScript object spread is evaluated left-to-right, `merged`'s `action` entry overwrites the `enum` discriminator. The emitted schema for `jira_project` has `action: { type: "string" }` with no `enum` — the JSON Schema gives no signal about valid action names.

Fix: add `if (key === "action") continue;` at the top of the inner loop (the discriminator always owns that key). The `list` action's Jira `action` parameter would also need renaming in `project.ts` to avoid it being silently omitted from the tool schema entirely.

---

**Everything else checked out:**

- The regression test (`expect(schema.oneOf).toBeUndefined()`) directly covers the API-400 bug being fixed.
- `zodToProperties` and `zodRequired` are fully removed with no remaining callers.
- `actionsIn` helper in `filter.test.ts` correctly updated to read from `properties.action.enum`.
- `additionalProperties: false` + passthrough `ActionKeySchema` + per-action Zod validation in `dispatchTool` is sound — no `.strict()` schemas found in any v2 tool file, so extra fields from other actions won't cause spurious Zod failures.
- The token-reduction claim is directionally true for tools with shared fields (`issueIdOrKey` appears in 6 of 9 `jira_issue` branches vs. once in the flat schema), though the description bloat partially offsets it.

src/tools/v2/dispatcher.ts
```diff
@@ -47,31 +47,118 @@ export interface ConsolidatedTool {
...
+      // dispatcher strips `action` from rawArgs before Zod
+      // validation, so an action schema declaring its own `action`
+      // field would be unreachable anyway. (jira_project.list has
+      // an `action` query param; this is the path that lets it
+      // coexist with the discriminator. The list action loses
+      // visibility of that param in the JSON Schema; agents can
+      // still pass it, additionalProperties:false is what blocks
+      // it — see below.)
+      if (key === "action") continue;
+      const js = zodFieldToJsonSchema(sub);
```
scottlepp, line 100: **Bug: `.optional().describe("..."\)` field descriptions are silently dropped from the emitted JSON Schema**

`zodFieldToJsonSchema` (called here) unwraps the optional wrapper first, then reads `.description` from the inner type (line 209). But Zod 4 attaches `.describe()` to whatever node it was called on — so `z.string().optional().describe("..."\)` puts the description on the `ZodOptional` wrapper, not on the inner `ZodString`. After `unwrap()`, the description is lost.

Verified via Zod 4 introspection:
```ts
const f = z.string().optional().describe("Comma-separated field list");
f.description           // "Comma-separated field list"  <- on wrapper
f._def.innerType.description  // undefined                <- inner has none
```

This affects 14 fields across 7 tool files currently in this PR:
- `issue.ts`: `fields` and `expand` (both `z.string().optional().describe(...)`)
- `search.ts`: `fields`
- `user.ts`: `query` (search), `project` and `issueKey` (assignable)
- `worklog.ts`: `startedAfter`, `timeSpent`, `comment`, `started`
- `board.ts`: `type`
- `sprint.ts`: `state`
- `filter.ts`: `groupId`
- `group.ts`: `permissions`

Fix: read description from the schema before unwrapping:
```ts
function zodFieldToJsonSchema(schema: ZodType): unknown {
  const description = (schema as unknown as { description?: string }).description; // before unwrap
  const inner = unwrap(schema);
  // ...
}
```

The sweep test does not catch this — it only asserts no top-level oneOf and non-empty action enum, not field description presence.

scottlepp:
Follow-up review after commit 3091722.

**Both prior blockers are resolved.**

Blocker 1 (field-type collision): the new per-field collection + property-level oneOf is correct. For jira_issue.fields, the emitted schema will be `{ oneOf: [{ type: "string" }, { type: "object", additionalProperties: true }] }`. The deepEqual dedup correctly collapses the three identical record variants (create/update/transition) to one branch. The helper itself is correct — I tested the same-count-different-keys edge case and it returns false as expected.

Blocker 2 (discriminator clobber): the spread order reversal (`...merged` first, discriminator last) combined with the `if (key === "action") continue` guard is correct. jira_project.list's `action` query param is now silently skipped from both the property bag and the per-action description suffix. The discriminator enum survives intact.

**additionalProperties: true** — no v2 action schema uses `.strict()` (grepped all files under `src/tools/v2/`), so unknown fields passed by an agent will be silently stripped by Zod's default behavior, not rejected. The reasoning in the commit message is accurate.

**sweep test** — the new describe("buildInputSchema across all consolidated tools") test would have caught the discriminator clobber (enum assertion) but not the field-type collision (first-wins was silent, no schema-shape signal). This matches the limitation noted in the prompt. The regression tests for both cases are in place.

**One new finding** posted as a line comment: the `.optional().describe("..."\)` pattern drops field descriptions silently. zodFieldToJsonSchema reads description from the unwrapped inner type, but Zod 4's .describe() attaches to the outer ZodOptional. 14 fields across 7 tool files are affected. The fix is a one-liner (read description before unwrapping). Not a blocker for correctness — agents still get the right type info — but the per-field documentation hints that are supposed to supplement the flat schema are silently missing for those fields.

**jira_project.list action param** — worth noting as a known tradeoff: the `action` query filter param on `project.list` is now invisible in both the JSON Schema and the per-action description suffix (the continue on line 99 skips it from both). The commit message acknowledges this. With additionalProperties: true an agent can still pass it, but only if it knows to from context outside the schema.

PR #196, opened 2026-05-05
v2: keep content in classic-mode summaries + add full-bypass flag

scottlepp:
## Summary
The v2 trim was designed for code-api mode, where the full Jira response is sandboxed to disk and the trim only runs on the in-band summary. Classic mode has no sandbox — the trimmed shape *is* the response — so any content the trim drops is permanently lost to the agent. Two fixes:

1. **Lighten `issueSummary`** (the most-used trim).
   - Drop the description cap (was 500 chars, now full body)
   - Drop the comment cap (was 300 chars, now full body) and inline every comment instead of just the last 3
   - Add `subtasks`, `parent`, and `issuelinks` with direction-specific relationship phrasing (`blocks` / `is blocked by`)
   - Field renames: `descriptionPreview` → `description`, `recentComments` → `comments`, `comment.preview` → `comment.body`

2. **Add a `full: true` meta-arg** on every consolidated tool.
   - Skips the trim, returns the raw Jira API response
   - Wired in the dispatcher (peeled off before per-action Zod validation), so every action with a trim gets the escape hatch automatically
   - Surfaced as a `boolean` property in each tool's input schema and mentioned in the description so agents can discover it

Other trims (`paginatedListSummary`, `bareListSummary`, `watcherList`, `voteList`) are unchanged — they still summarize for token reasons. `full: true` is the escape hatch for those.

## Why now
Repro: `jira_issue.get` for any issue with a multi-paragraph description / linked tickets / subtasks returned a 400-char preview, dropped `issuelinks` and `subtasks`, and capped each of the last 3 comments at 300 chars. The agent had no way to recover the missing content. After this PR, the same call returns the full description, every comment, every issuelink, and every subtask.

## Schema-breaking changes
The `IssueSummary` field set changed:
- `descriptionPreview` (string, capped) → `description` (string, full)
- `descriptionTruncated` (boolean) → removed
- `recentComments` (last 3, each preview-capped) → `comments` (all, full body)
- `comment.preview` → `comment.body`
- `+parent`, `+subtasks`, `+issuelinks` (new fields)

## Test plan
- [x] `npx vitest run` — 441/441 passing
- [x] `npx tsc --noEmit` — clean
- [x] `npm run build` — clean
- [x] Updated `commentSummary`, `issueSummary` unit tests to cover full-body paths and new fields (issuelinks direction, subtasks, parent)
- [x] New dispatcher tests: trim applied by default, `full: true` bypasses, `full: false` is identity, `full` not forwarded to Jira
- [x] New schema test: every tool exposes a `full` boolean in its input schema
- [ ] Manual: run against MON-694 with the updated server, confirm full description + linked tickets + subtasks come through. (Requires restarting the Claude client with `github:scottlepp/jira-mcp#pr/v2-trim-keep-content`.)

### DIFF
--- a/src/core/trim.ts
+++ b/src/core/trim.ts
@@ -146,6 +146,11 @@ function issueLinkSummary(l: {
   inwardIssue?: { id: string; key: string; fields?: { summary: string } };
   outwardIssue?: { id: string; key: string; fields?: { summary: string } };
 }): IssueLinkSummary {
+  // Jira's contract is exactly one of inwardIssue/outwardIssue per
+  // link record — the API returns each link once from each side, not
+  // a combined record with both. So presence of inwardIssue means
+  // *this* issue is on the outward side of the link type and the
+  // linked issue is the inward target, and vice versa.
   const direction: "inward" | "outward" = l.inwardIssue ? "inward" : "outward";
   const linked = l.inwardIssue ?? l.outwardIssue;
   return {
--- a/src/core/trim.ts
+++ b/src/core/trim.ts
@@ -16,14 +16,9 @@ import type {
   JiraUser,
 } from "../types/jira.js";
 
-const DESCRIPTION_PREVIEW_CHARS = 500;
-const COMMENT_PREVIEW_CHARS = 300;
-const RECENT_COMMENT_COUNT = 3;
-
-function truncate(s: string, max: number): string {
-  if (s.length <= max) return s;
-  return `${s.slice(0, max)}…`;
-}
+// (Description and comment caps were removed — see issueSummary.
+// Classic mode has no sandbox-on-disk fallback, so any cap silently
+// drops content the agent can't recover.)
 
 // --- User ---------------------------------------------------------------
 
@@ -51,7 +46,7 @@ export interface CommentSummary {
   author: UserSummary | null;
   created: string;
   updated: string;
-  preview: string;
+  body: string;
 }
 
 export const commentSummary = (c: JiraComment): CommentSummary => ({
@@ -59,7 +54,7 @@ export const commentSummary = (c: JiraComment): CommentSummary => ({
   author: userSummary(c.author),
   created: c.created,
   updated: c.updated,
-  preview: truncate(adfToPlainText(c.body), COMMENT_PREVIEW_CHARS),
+  body: adfToPlainText(c.body),
 });
 
 // --- Attachment ---------------------------------------------------------
@@ -96,6 +91,31 @@ export const projectSummary = (p: JiraProject): ProjectSummary => ({
 
 // --- Issue --------------------------------------------------------------
 
+export interface IssueLinkSummary {
+  id: string;
+  type: string;
+  // "inward" or "outward" — the direction relative to *this* issue.
+  direction: "inward" | "outward";
+  // The semantic relationship from this issue's POV (e.g. "blocks",
+  // "is blocked by"). Jira stores both phrasings on the link type;
+  // surfacing the one that matches `direction` lets the agent read
+  // the relationship without consulting docs.
+  relationship: string;
+  issue: { key: string; summary?: string };
+}
+
+export interface SubtaskSummary {
+  id: string;
+  key: string;
+  summary: string;
+}
+
+export interface ParentSummary {
+  id: string;
+  key: string;
+  summary?: string;
+}
+
 export interface IssueSummary {
   key: string;
   id: string;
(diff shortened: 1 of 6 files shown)


tests/tools/v2/dispatcher.test.ts
```diff
@@ -132,6 +147,55 @@ describe("dispatchTool", () => {
...
+    const ctx = makeMockClient();
+    ctx.setReturn([]);
+    await dispatchTool(fixtureTool, fixtureManifest, ctx.client, {
+      action: "list",
+      full: true,
+    });
+    expect(ctx.calls[0]).toMatchObject({ api: "get", path: "/x" });
+    const query = (ctx.calls[0] as { query?: Record<string, unknown> }).query;
+    expect(query ?? {}).not.toHaveProperty("full");
+  });
```
scottlepp, line 221: **Missing test: `full: true` on an action with no trim is a no-op**

Every new `full` test exercises `fx.list`, which has `trim: "bareList"`. There is no test for calling `full: true` on `fx.get` or `fx.create` (neither has a trim). The dispatcher always takes the `invokeOperationRaw` path when `wantFull` is true, so for an untrimmed action it returns the raw response just as `invokeOperation` would — correct, but untested.

Without a test, a future refactor that checks `op.trim` before routing could accidentally change behavior for untrimmed actions. Add a case like:

```ts
it("full: true on an untrimmed action returns the same result as without full", async () => {
  const ctx = makeMockClient();
  const raw = { id: "42", name: "thing" };
  ctx.setReturn(raw);
  const withFull = await dispatchTool(fixtureTool, fixtureManifest, ctx.client, { action: "get", id: "42", full: true });
  ctx.setReturn(raw);
  const withoutFull = await dispatchTool(fixtureTool, fixtureManifest, ctx.client, { action: "get", id: "42" });
  expect(withFull).toEqual(withoutFull);
});
```

src/core/trim.ts
```diff
@@ -107,18 +127,44 @@ export interface IssueSummary {
...
 }
 
+function issueLinkSummary(l: {
+  id: string;
+  type: { name: string; inward: string; outward: string };
+  inwardIssue?: { id: string; key: string; fields?: { summary: string } };
+  outwardIssue?: { id: string; key: string; fields?: { summary: string } };
+}): IssueLinkSummary {
+  const direction: "inward" | "outward" = l.inwardIssue ? "inward" : "outward";
+  const linked = l.inwardIssue ?? l.outwardIssue;
```
scottlepp, line 155: **Correctness edge case: issuelink with both `inwardIssue` and `outwardIssue` populated**

The Jira REST API contract says a link has exactly one side populated, but `JiraIssueLink` in `src/types/jira.ts` declares both as optional without any mutual-exclusivity guarantee. The function here picks `direction = "inward"` whenever `l.inwardIssue` is truthy and silently discards `l.outwardIssue` via the `??` chain:

```ts
const direction: "inward" | "outward" = l.inwardIssue ? "inward" : "outward";
const linked = l.inwardIssue ?? l.outwardIssue;
```

If Jira ever returns a malformed link with both sides set (some Server versions are known to do this on `Cloners`-type links), the outward issue is silently ignored. This is a low-likelihood but silent data loss. A defensive guard like:

```ts
if (l.inwardIssue && l.outwardIssue) {
  // Jira shouldn't send both; log and prefer inward.
}
```

or at least a comment explaining the assumption would make the intent explicit.

scottlepp:
**1 blocker, 2 suggestions.**

**Blocker**

- **`docs/MIGRATION.md` lines 73–76 — stale field names in the documented response shape.** The example JSON still shows `descriptionPreview`, `descriptionTruncated`, and `recentComments`. Those fields no longer exist after this PR. Programmatic consumers following this guide will wire up `result.descriptionPreview` / `result.recentComments` and silently get `undefined`. Since MIGRATION.md is the explicit contract reference for v2 consumers and the PR description calls these schema-breaking renames, the doc must be updated before merge. (MIGRATION.md is not in the diff, which is why this is a summary comment rather than a line comment.) The example block needs to reflect the new shape: `description` (full string), `comments` (array), no `descriptionTruncated`, plus the new `parent`, `subtasks`, and `issuelinks` fields. The sentence on line 82 that says "Field names are stable across versions" should also be revised or removed — this PR explicitly breaks that claim.

---

**Suggestions (non-blocking)**

1. **`src/core/trim.ts:150` — issuelink with both sides populated.** Flagged inline. Jira's contract says exactly one of `inwardIssue`/`outwardIssue` is populated per link, and the existing behavior (prefer inward) is reasonable. The concern is it's a silent drop with no comment explaining the assumption. Low real-world risk; a one-line comment suffices.

2. **`tests/tools/v2/dispatcher.test.ts` — missing test for `full: true` on an untrimmed action.** Flagged inline. All four new `full` tests exercise `fx.list` which has `trim: "bareList"`. There is no test asserting that `full: true` on `fx.get` (no trim) returns the same result as calling it without `full`. The behavior is correct by inspection of `invokeOperationRaw` vs `invokeOperation` when `op.trim` is undefined, but a regression test pins that invariant.

---

**Things checked and found clean:**

- No existing action schema in `src/tools/v2/` has a field named `full` — no collision risk with the new meta-arg.
- No references to `descriptionPreview`, `descriptionTruncated`, `recentComments`, or `comment.preview` remain in `src/` or `tests/` — all call sites were updated.
- `src/codeapi/bridge.ts` does not depend on field names — it passes the trim projection function through `trimRegistry` and uses `sandbox()` on the output. The rename is transparent to the code-api path.
- `tests/integration.test.ts` uses neither the old nor the new field names directly for `jira_issue.get` — no regression there.
- No Jira API operation parameter is named `full` in `src/core/operations.ts` — no shadowing risk.
- The dispatcher correctly strips `full` before passing args to Zod validation and before forwarding to Jira — verified in the diff.
- `IssueLinkSummary` having both `direction` and `relationship` is intentional and useful: an agent reading "direction: outward, relationship: blocks" gets a human-readable sentence without needing to know Jira's inward/outward naming convention. Not redundant.
- README line 78 says "recent comments" in prose but doesn't reference field names by string — stale phrasing but not a contract issue.
- The unchecked manual test item (MON-694) is noted; no live creds available in this environment.

PR #210, opened 2026-05-13
Migrate to @scottlepper/mcp-toolkit (v2.1.0)

scottlepp:
## Summary

Migrate jira-mcp's generic plumbing onto [@scottlepper/mcp-toolkit](https://github.com/scottlepp/lean-mcp-toolkit) (extracted from this repo + confluence-mcp). 19 files changed, **-1677 net LOC**. Cuts **v2.1.0**.

Toolkit work that landed first: v0.3 (retry+pool, LRU, streaming, disk-cache, `full` + `ToolError`), v0.4 (`.tool` field on `DispatchError`).

## What moved to the toolkit

| jira-mcp file | Replacement |
|---|---|
| `src/core/sandbox.ts` | `createSandbox({rootName, sessionEnvVar})` — kept as thin Jira config |
| `src/core/manifest.ts` | Toolkit `manifest` + Jira-specific `isAgile` extension, agile-routing `executeJiraOp`, and `JIRA_DISABLED_ACTIONS`-aware `assertOperationEnabled` |
| `src/core/trim.ts` lower helpers | Toolkit `trim` (Jira projections stay) |
| `src/core/http.ts` | **Deleted** — direct imports from `@scottlepper/mcp-toolkit/transport` |
| `src/core/lru.ts` | **Deleted** — direct imports from `@scottlepper/mcp-toolkit/lru` |
| `src/core/attachments.ts` streaming | Toolkit `streaming` (issue-key validation, preview, size verify stay) |
| `src/core/tenant-cache.ts` disk I/O | Toolkit `disk-cache` (Jira hostKey wrapper stays) |
| `src/codeapi/bridge.ts` | Toolkit `bridge` with Jira executor injection + `jmcp-` socket prefix |
| `src/codeapi/boot.ts` | Thinned to orchestrate cleanup → CLI resolve → startBridge → set socket env |
| `src/codeapi/tool.ts` | Toolkit `createCodeApiTool` definition + Jira-flavored `buildResponse` |
| `src/tools/v2/dispatcher.ts` | Toolkit `dispatch` (jira-permissive `buildInputSchema` stays — needed for `project.list`'s `action` query param) |
| `src/cli/index.ts` argv/help | Toolkit `createCli` |
| `src/index.ts` transport/signal wiring | Toolkit `startStdioServer` |

## What stayed in jira-mcp

Anything vendor- or product-specific: `auth/jira-client.ts`, `core/operations.ts`, `core/trim-registry.ts`, `core/adf.ts`, `resources/`, all 16 `tools/v2/*.ts`, `cli/direct.ts`, `cli/install-skill.ts`. The "lean" toolkit boundary is vendor-agnostic runtime primitives only.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 417 tests pass (down from 435; -18 were redundant tests of toolkit-owned `http.ts`/`lru.ts` re-exports)
- [ ] `npm run test:integration` with real Jira creds
- [ ] `npm run benchmark` — confirm tool-list bytes unchanged for classic and code-api modes
- [ ] `npm run inspector` — `jira_issue action=get` envelope byte-identical to a pre-migration baseline
- [ ] Code-api mode end-to-end: confirm `JIRA_MCP_SOCKET` inherited by subprocess, `jira-cli issue.get` via bridge
- [ ] Graceful SIGTERM drains in order (transport → bridge → HTTP pool)


src/tools/v2/dispatcher.ts
```diff
@@ -264,110 +274,53 @@ function zodFieldToJsonSchema(schema: ZodType): unknown {
...
+      },
     );
+    return result;
   } catch (err) {
+    // Re-shape the toolkit's `DispatchError` / manifest's
+    // `OperationError` as `ToolError` so callers can match on the
+    // tool-shaped class and read `.tool` directly. (Toolkit
+    // `dispatch()` always throws the base `DispatchError`; the
+    // `ToolError` subclass is opt-in.)
+    if (err instanceof DispatchError) {
```
scottlepp, line 318: The toolkit's `dispatch()` enforces that `full: true` is only valid for GET-verb actions (see toolkit `dispatcher.js` ~line 193: it throws `DispatchError` if `op.verb !== "GET"`). The old local dispatcher had no such restriction — it called `invokeOperationRaw` for any verb when `wantFull` was true. An agent that passes `full: true` to a mutation action (e.g. `issue.create`, `comment.add`) would have received the raw POST/PUT response before; now it gets a `DispatchError` re-wrapped as `ToolError`.

The PR description flags the `full: true` behavior as a frozen v2 contract. If mutation actions are intentionally excluded (they were never documented as supported), add an explicit call in this catch block to re-throw the verb-restriction `DispatchError` with a clear message, or confirm in the PR description that GET-only is the intended semantic.

src/core/attachments.ts
```diff
@@ -249,48 +164,42 @@ export async function downloadAttachment(
...
+      transport: opts.transport,
+    });
   } catch (err) {
-    await fs.rm(tempPath, { force: true });
+    // Rewrite the toolkit's generic "Download failed for <url>" into
+    // the v2 message shape so any error monitors keep matching. The
+    // toolkit's message already includes the HTTP code and body
+    // prefix — we just want the attachment id + filename out front.
+    const msg = err instanceof Error ? err.message : String(err);
+    const httpMatch = /HTTP (\d+) (.*)$/.exec(msg);
```
scottlepp: The toolkit throws: `Download failed for ${url}: HTTP ${code} ${body.slice(0,200)}`. The regex `/HTTP (\d+) (.*)$/` captures everything after the status code via `(.*)`, but `.` does not match newlines in JavaScript. If the HTTP error body contains a newline, the capture group stops at the first `\n`, silently truncating the error detail in the rethrown message. Not a correctness blocker (the error is still thrown and the Jira request is rejected), but the rethrown message may be less useful than expected when the error body is multi-line JSON. Consider `[\s\S]*` or `/HTTP (\d+) ([\s\S]*)$/` if preserving multi-line bodies matters.

scottlepp:
One suggestion, one nit.

**Suggestion — `src/tools/v2/dispatcher.ts` line 318:** The toolkit's `dispatch()` restricts `full: true` to GET-verb operations and throws `DispatchError` for any other verb. The old local dispatcher had no such restriction and would call `invokeOperationRaw` on any verb. The PR description lists `full: true` behavior as a frozen v2 contract. Whether mutation actions with `full: true` were ever legitimately used is unclear, but the behavior change is unacknowledged. Either document that `full: true` is GET-only by design, or add a note in the catch block that the verb-restriction error is intentional.

**Nit — `src/core/attachments.ts` line 188:** The error-message rewrite regex `(/HTTP (\d+) (.*)$/)` uses `.` which does not match newlines. If a Jira error response body contains `\n`, the captured body is silently truncated at the first newline. The error is still thrown and the file is not written, so this is not a correctness bug — but the rethrown message may omit useful context when the error body is multi-line JSON. `[\s\S]*` would capture the full body.

Everything else reviewed cleanly:
- `makeJiraExecutor` / `assertOperationEnabled` propagation is consistent across all three call paths (bridge, dispatcher, direct).
- `ToolError` constructor argument order is correctly adapted to the new `(message, action, tool?)` signature.
- `defaultBridgeAddress()` calls `jiraSandbox.sessionCacheDir()` at invocation time, not module-load time — correct.
- `defaultCliPath()` resolves `build/codeapi/boot.js → ../cli/index.js = build/cli/index.js`, pointing at jira-mcp's own CLI, not the toolkit.
- `startStdioServer` awaits the `onShutdown` callback before `process.exit(0)`, so the ordered shutdown (transport → bridge → HTTP pool) is guaranteed.
- The tenant-cache format change (`{cloudId,fetchedAt}` → `{v,fetchedAt}` + sha256 filename) is documented and backward-compatible (old entries are orphaned misses, not parse errors).
- No remaining imports of the deleted `src/core/http.ts` or `src/core/lru.ts`.

